Issue #33, opened 2026-06-20
Claude Code plugin install fails: build/index.js missing (build/ is gitignored, not produced on clone)


oliverames:
## Summary

When `apple-notes-mcp` is installed as a **Claude Code plugin** (via the marketplace, which installs by git clone rather than from npm), the server fails to start because the compiled `build/index.js` is never present. The `.mcp.json` launches `node ${CLAUDE_PROJECT_DIR}/build/index.js`, but `build/` is git-ignored, so a clean clone never contains it, and the plugin install does not reliably produce it.

## Symptom

Claude Code reports the server as **"Failed to connect."** The MCP log shows:

```
Error: Cannot find module '.../apple-notes-mcp/apple-notes/2.0.1/build/index.js'
  code: 'MODULE_NOT_FOUND'
```

The process exits in roughly 90ms, so the connection closes immediately with `MCP error -32000: Connection closed`.

## Root cause

The repo serves two distribution channels that disagree about where the runnable code lives:

- **npm**: `.npmignore` excludes `src/` and publishes the compiled `build/` directory, which is generated by the `prepare` lifecycle script. This path works.
- **Claude Code plugin**: the marketplace installs by git clone (the install record carries a `gitCommitSha`). `.gitignore` excludes `build/`, so the cloned tree contains `src/` but no `build/`. The plugin's `.mcp.json` then points at `build/index.js`, which does not exist.

Two compounding factors:

1. `build/` is git-ignored, so it is never part of a clone.
2. `"prepare": "husky && npm run build"` short-circuits on the `&&`. In a plugin-clone context `node_modules/` was installed (so `npm install` did run), yet `build/` was still missing, which is consistent with `husky` erroring before `npm run build` is reached.

## Reproduction

1. Install the plugin through the Claude Code marketplace.
2. Start Claude Code and check MCP status; the `apple-notes` server shows as failed.
3. Inspect the install path: `src/` and `node_modules/` are present, `build/` is absent.

## Workaround

Building in place restores the server:

```
cd <plugin-install-path>/apple-notes-mcp/apple-notes/<version>
npm run build
```

After that, `build/index.js` exists and the server completes the MCP `initialize` handshake normally. The fix does not survive the next plugin upgrade, since the cache is recreated from a fresh clone.

## Suggested fixes (any one resolves it)

- Commit the compiled `build/` output for the plugin distribution (or ship a prebuilt release artifact), so a clean git clone is runnable without a build step.
[...]

PR #34, opened 2026-06-20
Add Codex plugin marketplace alongside the Claude one

oliverames:
## What this does

This adds Codex support alongside the existing Claude Code plugin, so the same Apple Notes MCP server can be installed from Codex's plugin marketplace. The Claude side is unchanged; this sits next to it.

The new pieces:

- `.agents/plugins/marketplace.json`: the Codex marketplace, pointing at a local `./codex` plugin.
- `codex/`: the Codex plugin itself. Its `.mcp.json` runs the published server with `npx -y apple-notes-mcp`, which matches the `npx apple-notes-mcp` invocation you already document for Claude Desktop. It also bundles the same Apple Notes skill and a couple of display assets (the Notes icon and a screenshot).
- `scripts/sync-plugin-version.mjs`: keeps the Claude and Codex manifests on the same version. It replaces the inline one-liner in the `version` script and now updates all four manifests on `npm version`.
- README and `package.json`: a short Codex install section, plus a `codex` keyword.

## One small fix that fell out of this

When I wired up the version-sync script and ran it, it caught that `.claude-plugin/marketplace.json` still listed the plugin at `1.1.1` while `package.json` and `.claude-plugin/plugin.json` are at `2.0.1`. I let the script correct it to `2.0.1`. That drift is exactly what the script is meant to prevent going forward.

## Why npx the published package

I pointed the Codex MCP config at `npx -y apple-notes-mcp` rather than building from source. It is the runtime you already recommend for Claude Desktop, and it uses the package you publish to npm, so Codex users get the compiled server without a local build step.

## Testing

- `npm run typecheck`, `npm run build`, and `npm test` (361 passing) all green.
- Every manifest validated as JSON and confirmed at `2.0.1`.

Happy to adjust the marketplace naming, category, or display copy to match whatever conventions you prefer.


sweetrb:
Thanks @oliverames! The `sync-plugin-version.mjs` script is a nice cleanup over the inline one-liner, and catching the `1.1.1` version drift was a good catch. The `npx` approach for the Codex MCP config is the right call — and as a bonus it sidesteps the build issue you filed in #33.

PR #35, opened 2026-06-20
Dedupe CoreData notes returned more than once by Notes.app

oliverames:
## The problem

When you ask Notes.app for all of an account's notes, it can hand back the same CoreData note more than once. That made `search-notes` and `list-notes` occasionally return duplicate entries for a single note.

## The fix

Dedupe on note ID at two layers, so a duplicate from either side is dropped:

- **AppleScript loop**: track seen IDs and skip a note whose ID has already been added before appending it to the result list.
- **JS parsing**: guard with a `Set<string>` so any duplicate that still slips through is filtered before results are returned.

This is applied consistently across the three code paths that enumerate notes: the search path, the date/limit list path, and the simple list path. The simple list path now uses a repeat loop (emitting name and ID) instead of `name of notes`, so it can dedupe the same way the other paths do.

## Tests

Two new tests:
- duplicate note IDs returned by Notes.app are deduped
- distinct notes that happen to share a title are preserved (dedupe is by ID, not title)

Full suite: 363 passing. `npm run typecheck` and `npm run build` are clean.

### DIFF
--- a/src/services/appleNotesManager.test.ts
+++ b/src/services/appleNotesManager.test.ts
@@ -695,6 +695,22 @@ describe("AppleNotesManager", () => {
       expect(results[1].folder).toBe("Notes");
     });
 
+    it("deduplicates duplicate note IDs returned by Notes.app", () => {
+      mockExecuteAppleScript.mockReturnValue({
+        success: true,
+        output: [
+          ["Not uploaded", "x-coredata://ABC/ICNote/p1", "Notes"].join(F),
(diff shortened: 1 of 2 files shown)


sweetrb:
Thanks @oliverames! The dual-layer dedup (AppleScript + JS Set) is the right defensive approach here, and the test that verifies same-title/different-ID notes are preserved is exactly the edge case worth covering.

PR #40, opened 2026-06-22
Add Hermes and Antigravity plugin packaging

oliverames:
Hi Rob,

Following on from the Codex marketplace in #34, here's packaging for two more MCP hosts: Hermes and Antigravity. I run apple-notes-mcp on both and have been carrying these manifests in my own fork, but they belong here next to the Claude and Codex ones instead of in a fork I keep rebasing.

What this adds:

- `.hermes-plugin/` with marketplace.json, plugin.json, and mcp.json
- `.antigravity-plugin/` with marketplace.json, plugin.json, mcp_config.json, and a copy of the apple-notes skill

Both launch the server with `npx -y apple-notes-mcp`, the same command codex/.mcp.json already uses. That keeps them on the published npm package, so there's no build step on clone and nothing new to install. The change is additive: it doesn't touch any of the existing Claude or Codex packaging, and it's macOS-only like everything else here.

One thing I left out on purpose. The `version` script in package.json git-adds the Claude and Codex manifests so their versions stay in sync on release. I didn't wire these two in, to keep the diff small, but I'm glad to add them in a follow-up if you'd like them kept in lockstep.

Thanks for taking the earlier fixes. Happy to restructure any of this to match how you'd rather lay it out.

Best,
Oliver

### DIFF
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -1,0 +1,36 @@
+{
+  "name": "apple-notes-mcp",
+  "metadata": {
+    "description": "Apple Notes integration marketplace for Claude Code, Codex, Hermes, and Antigravity.",
+    "version": "1.0.0"
+  },
+  "plugins": [
+    {
+      "name": "apple-notes",
+      "version": "2.1.0",
+      "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
+      "source": {
+        "source": "local",
+        "path": "."
+      },
+      "category": "productivity",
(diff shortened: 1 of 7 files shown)


sweetrb:
Merged — thanks @oliverames! Clean, additive packaging that mirrors the Codex layout exactly, and I appreciate you flagging up front that you'd left the two new manifests out of `sync-plugin-version.mjs` to keep the diff small. I wired `.hermes-plugin/` and `.antigravity-plugin/` into the sync script in a follow-up so their versions track `package.json`, and added README install notes + a CHANGELOG entry. We're now bringing the same Hermes/Antigravity packaging to the other Apple MCP servers (mail, numbers, photos) for parity.

PR #44, opened 2026-06-23
feat: reveal folders, accounts, and attachments in Notes.app

oliverames:
The Notes scripting dictionary's `show` command works on a note, folder, account, or attachment, but the server only wrapped it for notes (`show-note`). This adds the three that were missing and closes the "show or reveal a note, folder, account, or attachment" item under AppleScript Surface Gaps in TODO.md.

### What's new

`show-folder` reveals a folder by id, and `show-account` reveals an account by id. The ids come from `list-folders` and `list-accounts`.

`show-attachment` is the more involved one. Attachments are elements of a note, so AppleScript can't reach them by id at the application level the way it can a note or folder. It resolves the attachment inside its note first, using the same lookup `save-attachment` already does, then runs `show` on the matched reference. The inputs are `noteId` and `attachmentId`, matching `save-attachment` and `fetch-attachment`.

Each tool mirrors `show-note`: an optional `separately` flag, AppleScript built through the existing `buildAppLevelScript`, `sanitizeId`, and escape helpers, and a `false` return on failure instead of throwing.

### Backwards compatibility

Everything here is additive. No existing tool, signature, description, or handler behavior changed, so this stays within the "prefer adding new tools" and "all new parameters optional" guidance in the roadmap.

### Tests and docs

There are 10 new tests in `appleNotesManager.test.ts` covering the success path, the `separately` flag, the failure path, and the attachment-not-found case, all against mocked `runAppleScript`. README, CHANGELOG, and both skill manifests (`skills/apple-notes/SKILL.md` and `codex/skills/apple-notes/SKILL.md`) are updated.

One thing I left alone on purpose: the skill manifests were already missing `show-note`, `get-selected-notes`, and `get-default-location` from an earlier release, so those tables had drifted before this change. I added only the three new tools to keep the diff focused. The older gaps are worth a separate cleanup.

`npm run lint`, `npm run typecheck`, `npm test` (408 passing), `npm run build`, and `npm run format:check` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- **Reveal folders, accounts, and attachments in Notes.app.** Three new tools extend the existing `show-note` to the rest of the objects the Notes scripting dictionary's `show` command accepts: `show-folder` (by folder id), `show-account` (by account id), and `show-attachment` (by note id + attachment id, since attachments are note-scoped). Each takes an optional `separately` flag, mirroring `show-note`. This closes the "show or reveal a note, folder, account, or attachment" surface gap from the roadmap; everything is additive AppleScript, and no existing tool changed.
 
 ## [2.3.0] - 2026-06-23
 ### Added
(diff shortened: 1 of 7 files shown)


sweetrb:
Merged into v2.4.0 — thanks @oliverames. Nice job extending `show` to the rest of the objects the scripting dictionary accepts. The standout is `show-attachment`: since attachments aren't addressable at the application level, resolving the attachment inside its note first (the same lookup `save-attachment` uses) then running `show` on the matched reference is exactly the right approach. Clean mirror of `show-note` throughout, with proper id sanitization and the optional `separately` flag.

PR #45, opened 2026-06-23
test: lock Notes-normalized HTML to Markdown, and warn about attachments on update

oliverames:
The "Formatting and Note-Body Safety" item in TODO.md was mostly handled already by the skill's formatting guidance (#42). Two pieces were still open, and this covers both.

### Regression fixtures for normalized HTML

Apple Notes rewrites HTML when it saves a note. It wraps blocks in `<div>` instead of `<p>`, drops in `<div><br></div>` spacer rows, and hands back its own tag soup when you read the body. Nothing in the suite pinned what `getNoteMarkdown` does with that shape, so a change to the Turndown rules could quietly shift the output.

`src/services/__fixtures__/notesNormalizedHtml.ts` pairs representative normalized bodies with the Markdown the server produces today, and `notesHtmlMarkdown.test.ts` runs them through the real conversion path. These are characterization tests, so they lock current behavior rather than claim it's ideal. Two quirks are now written down where a reviewer will see them:

- A `<div><br></div>` spacer becomes a stray two-space line. That's the Markdown side of the whitespace-accumulation behavior CLAUDE.md already warns about.
- `<tt>` is dropped. The text survives, the code styling doesn't round-trip, even though the skill suggests `<tt>` for code in notes.

### Attachment warning on update-note

The skill's "Attachment-Safe Updates" section tells agents to run `list-attachments` before rewriting a note that might hold files, but the `update-note` tool description itself said nothing about it. An agent reading only the tool metadata wouldn't know a full-body replace can drop embedded images, scans, PDFs, or audio. The description and the README `update-note` section now say so. The skill was already correct; this just brings the MCP-visible text in line.

### Scope

Docs, tests, and one description string. No tool behavior, signature, or parameter changed. 7 new tests, and `npm run lint`, `npm run typecheck`, `npm test`, `npm run build`, and `npm run format:check` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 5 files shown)


sweetrb:
Merged into v2.4.0 — thanks @oliverames. The characterization fixtures are the right move: pinning the two known quirks (a `<div><br></div>` spacer leaving a stray two-space line, and `<tt>` being dropped so code styling doesn't round-trip) means a future Turndown change can't silently shift the Markdown output. And surfacing the attachment-loss warning in the `update-note` description itself — not just the skill — is the part an agent reading only tool metadata actually needs.

PR #46, opened 2026-06-23
feat: add get-note-plaintext tool

oliverames:
The scripting dictionary gives every note a read-only `plaintext` property, which is the body with all the HTML stripped out. The server never used it. `get-note-content` returns the raw HTML, and `get-note-markdown` runs that HTML through Turndown, but there was no way to just ask for the text.

`get-note-plaintext` reads that property directly, by id or title, the same shape as `get-note-content`. Because Notes computes the plaintext itself, it is more faithful than stripping the HTML on our end, and it skips the conversion pass.

### Why it's worth having

An agent that only wants the words (to summarize, search, or feed somewhere else) currently has to pull the HTML and clean it up, or take the Markdown and accept the markup. This gives a clean text read with no conversion in the middle.

### Details

- New `getNotePlaintext(title, account)` and `getNotePlaintextById(id)` on the manager, built with the existing `buildAccountScopedScript` / `buildAppLevelScript` / `sanitizeId` / escape helpers, returning an empty string on failure like their `get-note-content` siblings.
- Password-protected notes are checked first and get the same "unlock it in Notes.app" message as `get-note-content`.
- 6 new tests covering the title and id paths, account scoping, the failure path, and the malformed-id guard.
- README, CHANGELOG, and both skill manifests updated.

No existing tool, signature, or behavior changed. `npm run lint`, `npm run typecheck`, `npm test`, `npm run build`, and `npm run format:check` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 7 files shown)


sweetrb:
Merged into v2.4.0 — thanks @oliverames. Reading the note's native `plaintext` property is the right call: it's more faithful than stripping HTML on our end and skips the Turndown pass entirely, giving a clean third read shape alongside `get-note-content` (HTML) and `get-note-markdown`. The password-protected pre-check matching `get-note-content` keeps the behavior consistent too.

PR #47, opened 2026-06-23
docs: App Intents bridge feasibility and verified read-only SQLite columns

oliverames:
`TECHNICAL_NOTES.md` was last touched in December 2024, and a few of the roadmap's open questions have real answers now. I spent a session researching them, verified against this Mac (macOS 27, Notes 4.13), the live `shortcuts` CLI, and Apple's developer docs, and wrote the findings down so the next person doesn't have to re-derive them.

### The App Intents question has a hard answer

The roadmap's "App Intents Bridge" lists "native Swift helper that invokes App Intents directly" as a possible approach. It isn't possible. App Intents are system-invoked and app-local: an app exposes its own actions, and only the system (Siri, Spotlight, Shortcuts) calls them. There is no public way to `perform()` another app's intents, so a helper would just fall back to AppleScript.

The one route that does work is `shortcuts run` against a Shortcut the user has installed. That's real, but it is a BETA, opt-in path: it needs an active GUI login session (no SSH or launchd), a one-time manual install, and it is plain text only. The new section lays out which capabilities that bridge reaches (pin/unpin, tags, move to folder, append checklist item, append text, attach a file) and which stay GUI-only no matter what (prepend, Markdown body, checklist toggle, tables, note links).

### The read-only SQLite metadata is closer than the roadmap implies

The "Read-Only SQLite Metadata Expansion" item reads as mostly protobuf work. A lot of it is not. The highest-value fields are plain scalar columns on `ZICCLOUDSYNCINGOBJECT`, readable with an ordinary SELECT: `ZISPINNED`, `ZHASCHECKLIST` / `ZHASCHECKLISTINPROGRESS`, `ZISRECOVERINGFROMTRASH`, `ZSMARTFOLDERQUERYJSON`, the snippet fields, and the password hint. I verified these against a live store on macOS 27. `ZISPINNED` is worth calling out: AppleScript has no pinned property, so reading that column is the only way to know a note's pin state, which closes the read half of the "pinned notes" limitation the project documents as impossible.

### What's here

Docs only. A new App Intents section and a verified-columns subsection in `TECHNICAL_NOTES.md`, both cross-linked from annotations in `TODO.md`, plus the date bump. No code.

I am planning to follow this with the read-only metadata tool as the first implementation step, since it is the lowest-risk piece.

### DIFF
--- a/TECHNICAL_NOTES.md
+++ b/TECHNICAL_NOTES.md
@@ -130,6 +130,29 @@ for pk, data, title in cursor:
 - **Quit Notes**: For consistent reads, quit Notes.app first
 - **Full Disk Access**: Required to access the Group Containers path
 
+### Read-Only Metadata Columns (verified macOS 27 / Notes 4.13)
+
+Most useful note metadata lives as plain scalar columns on `ZICCLOUDSYNCINGOBJECT`, so
+it can be read with an ordinary `SELECT` and needs no protobuf decoding, unlike the
+`ZICNOTEDATA.ZDATA` body blob. The columns below were confirmed against a live store on
+macOS 27.0 (Notes 4.13). Column names and availability shift between OS releases, so
+treat them as version-specific and feature-detect with `PRAGMA table_info` before use.
+
+| Column | Exposes | Why it matters |
+|--------|---------|----------------|
+| `ZISPINNED` | Pinned state (boolean) | AppleScript has no `pinned` property, so this is the only read path for pin state |
+| `ZHASCHECKLIST`, `ZHASCHECKLISTINPROGRESS` | Whether a note has a checklist, and whether any item is still unchecked | Cheap flags without decoding the body |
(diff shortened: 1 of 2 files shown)


sweetrb:
Merged — thanks @oliverames, this is exactly the kind of feasibility writeup that saves the next person a whole research session. Two findings stand out:

1. The hard "no" on a Swift App Intents helper — confirming intents are system-invoked and app-local with no public cross-app `perform()`, so a helper would just fall back to AppleScript with a *worse* permissions story. That retires a tempting-but-dead-end roadmap item.
2. `ZISPINNED` as a plain scalar column on `ZICCLOUDSYNCINGOBJECT` — that closes the **read** half of the pinned-notes limitation we'd documented as impossible (AppleScript has no `pinned` property).

The `PRAGMA table_info` feature-detection over `sw_vers` gating, and the copy-then-read-only safety reminder, are the right calls. Looking forward to the read-only metadata tool as the low-risk first implementation step.

PR #48, opened 2026-06-23
feat: get-note-metadata (BETA) for pinned/trash/snippet via the NoteStore DB

oliverames:
AppleScript cannot read a note's pinned state. TECHNICAL_NOTES and the README both list it as impossible. It turns out the information is right there in the NoteStore database as a plain column, sitting next to the checklist data this server already reads, so it never needed protobuf work or App Intents.

`get-note-metadata` (BETA, by id) reads the columns AppleScript leaves out: `pinned`, `hasChecklist` and `hasChecklistInProgress`, `recoveringFromTrash`, `passwordProtected`, `passwordHint`, `snippet`, `widgetSnippet`, and `smartFolderQuery`.

### How it stays safe

The database is opened read-only and never written. I used `execFileSync` with an argument array rather than a shell string, so the path's spaces need no quoting and nothing is interpreted by a shell. The only dynamic value in any query is the note's primary key, which is constrained to digits before it goes anywhere near SQLite.

### Why it is BETA

The NoteStore schema changes between macOS versions. Rather than hardcode a column list, the reader asks `PRAGMA table_info` which columns exist and selects only those, so it returns fewer fields on an older or newer system instead of erroring. The BETA label sets that expectation. It also reads straight from the database with no AppleScript pre-check, which is deliberate: it can resolve trashed and recovering notes that `note id ...` can no longer find.

### Verification

Six unit tests cover the pk guard, column feature-detection, boolean coercion, the not-found path, and the Full Disk Access denial, all against a mocked database. I also ran the built reader against the real NoteStore on macOS 27, and it read `pinned: true` off an actually-pinned note, which is the headline AppleScript cannot do. Full suite green, plus lint, typecheck, build, and format checks.

This makes pinned state readable for the first time. It still cannot be set; that needs the Shortcuts bridge described in the App Intents research (#47). The README "Known limitations" note is updated to say exactly that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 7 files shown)


sweetrb:
Merged into v2.5.0 — outstanding work, @oliverames. This makes a note's **pinned** state readable for the first time, which we'd documented as impossible. The safety engineering is exactly right: the pk is digit-constrained (`/\/p(\d+)$/`) before it touches SQL, `execFileSync` with an arg array (no shell), the column list is a fixed allowlist, the DB is opened `-readonly`, and `PRAGMA table_info` feature-detection degrades gracefully across macOS schema changes instead of throwing. Reading straight from the DB (no AppleScript pre-check) so it resolves trashed/recovering notes is a nice touch, and the BETA label + the README 'readable but not settable' note set the right expectations.

PR #49, opened 2026-06-23
docs(skill): port four field-tested Apple Notes techniques into the skill

oliverames:
Working from a personal Apple Notes skill, I pulled four techniques into the repo's bundled skill that were not there yet. All four are generic Notes behavior, nothing tied to a particular setup.

1. **Verify with get-note-plaintext.** The Verification section already notes that stored HTML can look wrong while rendering fine. Now that `get-note-plaintext` is in (#46), it is the fastest way to check: it returns the plain text Notes itself shows, so you can confirm the title and line breaks survived without reading through normalized HTML.

2. **No decorative separators.** Horizontal rules, repeated dashes, and box-drawing characters render inconsistently in Notes. Added a line in Formatting Guidance to use a `<div><br></div>` spacer instead.

3. **maxBuffer as an attachment signal.** A `stdout maxBuffer length exceeded` error usually means the note has large embedded attachments. Added it to the existing list of attachment-risk tells, so an agent treats it as a reason to run `list-attachments` before a full-body replace.

4. **Optional title control without a duplicate.** This one is advanced, so it is written as optional with the default ("pass `title` only") kept front and center. If you do want to control the title HTML and hit the duplicate small-title line, you can create with a styled `<h1>` and then `update-note` with `newTitle: " "` so Notes uses the first body line as the sidebar title. The catch is documented too: after blanking the metadata title, the note's CoreData id can stop resolving, so you address it by display title afterward.

Both the canonical skill and the Codex mirror are updated, and the edited sections are byte-for-byte identical between them. Docs only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Documentation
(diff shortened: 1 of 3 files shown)


sweetrb:
Merged into v2.5.0 — thanks @oliverames. All four are solid, generic Notes guidance. The `stdout maxBuffer length exceeded` -> attachment-risk signal is a great practical tell (an agent won't otherwise connect that error to 'run list-attachments first'), and the optional title-control technique (styled `<h1>` + `update-note newTitle: ' '`) is handled well — kept clearly optional/advanced with its CoreData-id-resolution caveat documented.

PR #41, opened 2026-06-22
Expose additional AppleScript Notes surfaces

oliverames:
## Summary

- add `show-note`, `get-selected-notes`, and `get-default-location` tools
- enrich AppleScript-derived account, folder, and attachment metadata with IDs, default-folder data, shared state, URLs, and timestamps where available
- keep legacy parser compatibility for older unit fixtures and downstream mocks that still return plain account names or tab-delimited folders
- document the new tool surface in the README

## Verification

- `npm test` (390 tests)
- `npm run build`
- `npm run lint`
- `git diff --check`
- pre-push hook: `npm run typecheck` and `npm test`

PR #50, opened 2026-06-23
docs: mark completed items on the roadmap

oliverames:
Now that several of these have landed, the roadmap should show it. I added a one-line Status under each item that moved, plus a pointer in the header. Nothing was removed, so the original plan stays visible.

- **AppleScript Surface Gaps: Complete.** Shipped across #41 (reveal note, selected IDs, default location, account/folder metadata, richer attachment fields) and #44 (reveal folder, account, attachment). `get-note-plaintext` (#46) is a bonus on top.
- **Formatting and Note-Body Safety: Complete.** Safe-HTML and attachment-safe guidance in the skill (#42), the `list-attachments`-before-replace warning and Notes-normalized-HTML regression fixtures in #45.
- **Read-Only SQLite Metadata: Partially shipped (BETA).** `get-note-metadata` (#48) reads the scalar columns (pinned, checklist flags, trash/recovery, smart-folder query, snippets, password hint). The protobuf tier (OCR, summaries, attachment dimensions, transcripts) is still open.
- **App Intents Bridge: Researched, deferred.** The feasibility note already in that section explains why; the Status line just records the decision not to build it for now.

Docs only, `TODO.md` only.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,6 +1,6 @@
 # Apple Notes MCP - Improvement Roadmap
 
-*As of v2.1.0*
+*As of v2.1.0. Completion status updated 2026-06-23; see the Status line under each item.*
 
 Based on technical research into Apple Notes internals, the current
 AppleScript dictionary, Notes App Intents metadata, and analysis of other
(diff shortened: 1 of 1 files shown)


sweetrb:
Merged — thanks @oliverames. Keeping the roadmap honest as things land (Status lines, nothing removed so the original plan stays visible) is genuinely useful. The 'Read-Only SQLite Metadata: Partially shipped (BETA)' framing is accurate: get-note-metadata covers the scalar columns; the protobuf tier (OCR / summaries / attachment dimensions / transcripts) is still open.

PR #67, opened 2026-07-03
chore: exclude compiled tests from tracked build/

sweetrb:
Build-infra consistency fix. notes' tsconfig.json lacked the test-exclude that mail/numbers/photos have, so it was tracking 16 compiled `build/**/*.test.js` files.

Adds `"build"` and `"**/*.test.ts"` to the `exclude` array (matching the sibling repos) and removes the stale compiled test files so tracked `build/` is runtime-only.

No src change, no version bump (build-infra only — version-guard exempts non-src changes).


sweetrb:
Superseded by #66 (merged to main at 17:03 UTC), which made the identical change (tsconfig test-exclude + drop tracked build/**/*.test.js). Closing this redundant/conflicting PR; main already has runtime-only tracked build/.

PR #69, opened 2026-07-06
Bundle the entrypoint with esbuild so a bare clone runs without node_modules

### LINKED ISSUE #68: Marketplace clone still can't run the server: node_modules missing after re-clone (follow-up to #33)
Follow-up to #33. Committing `build/` fixed the missing-entrypoint half of the problem, but a marketplace clone still can't run the server, because the compiled output imports its runtime dependencies from `node_modules/` and a git clone never has those.

**What happened**

Claude Code's marketplace auto-update re-cloned the plugin from scratch tonight (confirmed in the clone's reflog: `clone: from https://github.com/sweetrb/apple-notes-mcp.git`). The fresh tree had the committed `build/`, but no `node_modules/`, so the server died on its first import at every session start:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@modelcontextprotocol/sdk'
imported from .../build/index.js
```

The plugin launches with `node ${CLAUDE_PLUGIN_ROOT}/build/index.js` (from `.claude-plugin/plugin.json`), so there is no install step anywhere in the path between "marketplace refresh" and "server process starts." apple-mail-mcp has the identical failure; I'm filing a companion issue there.

**Workaround**

Running `pnpm install && pnpm run build` in the clone restores it, verified with a piped JSON-RPC `initialize` handshake against `build/index.js`. I've added that step to my own maintenance scripts, but every marketplace user hits this on every refresh, and the failure mode is silent: the MCP server just never connects, and the reason only shows up in Claude Code's MCP logs.

**Proposed fix**

Bundle the entrypoint as a single dependency-free file so a bare clone runs with nothing but Node present:

```
esbuild src/index.ts --bundle --platform=node --format=esm --outfile=build/index.js
```

A `prepare`/`postinstall` script can't help here because the marketplace install path never runs a package manager. Bundling is the only approach that asks nothing of the host.

Seen on v2.5.7, macOS, Claude Code marketplace install.

oliverames:
Fixes #68.

## What changed

`npm run build` now typechecks with `tsc --noEmit` and then bundles `src/index.ts` into a single self-contained `build/index.js` with esbuild. A bare git clone, which is exactly what Claude Code's marketplace auto-update produces, now launches the server with nothing but Node installed. Only the bundled entrypoint is tracked in git; the per-module compiled files are removed and stay ignored, so `npm run dev` output no longer shows up as untracked churn. `tsc-alias` is dropped because esbuild resolves the tsconfig path aliases natively, and the entry shebang survives bundling so `bin` still works.

## Why

Committing `build/` (#65) fixed the missing-entrypoint half of #33, but the compiled output still imported its runtime dependencies from `node_modules/`, which a clone never has. The marketplace refresh path never runs a package manager, so every auto-update killed the server at session start with `ERR_MODULE_NOT_FOUND` on `@modelcontextprotocol/sdk`, and the failure was silent unless you went digging in Claude Code's MCP logs.

## The one wrinkle

`src/index.ts` reads `../package.json` at runtime through `createRequire` to report its version, so the artifact is one file plus `package.json` rather than strictly one file. I left that alone: every distribution layout (git clone, npm package, marketplace install) ships `package.json` in that exact spot, and keeping `src/` untouched keeps this PR outside version-guard's shipped-code rule. If you'd rather inline the version at build time with an esbuild define, that takes a small src change plus a version bump, and I'm happy to follow up.

## Verification

- `lint`, `typecheck`, `test` (440 passing), `format:check`, and `build` are all green locally.
- Live probe of the real failure mode: fresh `git clone` of this branch into an empty directory, no install step, then a piped JSON-RPC session against `node build/index.js`. `initialize` returns `serverInfo` for apple-notes 2.5.7 and `tools/list` returns all 34 tools with nothing on stderr.

## Left out of scope

- No version bump, since nothing under `src/` changed. The fix matters most for the git and marketplace paths, which track main; whether to cut an npm release for it is your call.
- `package.json` declares `"types": "build/index.d.ts"`, which the build has never emitted (the old tracked `build/` had only `.js` files). Pre-existing, so I didn't touch it here.
- apple-mail-mcp has the identical failure (sweetrb/apple-mail-mcp#78). I can send the same change there if you want it.


oliverames:
Pushed one follow-up commit: the build now adds the same createRequire banner as the apple-mail companion (sweetrb/apple-mail-mcp#79). The notes dependency graph doesn't currently trip esbuild's dynamic-require ESM shim, but the mail repo's does (nodemailer and imapflow require Node builtins dynamically), and without the banner a routine dependency bump here could silently reintroduce the exact marketplace breakage this PR fixes. Keeping the two build scripts identical also keeps the repos easy to diff against each other. Re-verified the bare-clone handshake and tools/list after the change.

sweetrb:
Merged and shipped in **v2.5.8**. Same clean esbuild-bundle approach, @oliverames — verified standalone (MCP `initialize` from a bare-clone layout, no `node_modules`). Good call keeping `src/` untouched so it stayed outside the version-guard rule.

PR #70, opened 2026-07-06
Reliability: retry transient failures by default, timeout/retry env knobs, doctor check for ad-hoc signed Node (v2.5.8)

oliverames:
Three changes aimed at the ways this server actually flakes in daily use. Each one came out of a real failure I hit or measured on my own machine, not a hypothetical.

**Note on the base: this branch is stacked on #69**, because both PRs regenerate the committed `build/index.js` and would conflict if cut independently. If you merge #69 first, this PR reduces to its last commit. Happy to rebase if you'd like a different shape.

## 1. Transient AppleScript failures retry by default

The retry machinery is already in the codebase, with proper transient-error classification (busy / not responding / lost connection / timeout) and exponential backoff, but `DEFAULT_MAX_RETRIES = 1` means it never fires unless a caller opts in, and no tool call does. So a momentarily busy Notes.app fails the user's tool call instantly even though a 1s-later retry would have succeeded. The default is now 2 (one retry after 1s). Non-transient errors like "note not found" never retry, exactly as before, and `APPLE_NOTES_MCP_MAX_RETRIES=1` restores fail-fast.

## 2. Env knobs for the execution policy

`APPLE_NOTES_MCP_TIMEOUT_MS`, `APPLE_NOTES_MCP_MAX_RETRIES`, and `APPLE_NOTES_MCP_RETRY_DELAY_MS`, following the exact validation pattern of the existing `APPLE_NOTES_MCP_MAX_BUFFER` (shared through a new `envPositiveNumber` helper). The 30s timeout is right for most libraries but a full-library search over a large collection legitimately exceeds it, and until now the only recourse was editing source. Per-call options still win over env, and the config-file fallback covers hosts that strip `env`. Documented in the README table.

## 3. `doctor` checks the Node runtime's code signature

Your own `docs/NODE-RUNTIME-AND-TCC-PERMISSIONS.md` documents the failure: an ad-hoc signed Node (typically Homebrew's) gets a new cdhash on every update, so macOS TCC silently revokes its Automation and Full Disk Access grants, which users experience as random permission loss. The doctor now inspects `process.execPath` with `codesign` and warns with a pointer to that doc. I verified the check live against both binaries on my machine: Homebrew's node reports warn (`Signature=adhoc`), a Developer-ID node reports ok with its Team ID.

## Verification

- 448 tests pass (8 new: retry default, env-knob resolution and precedence, invalid-value fallback, env timeout in the error message, and three signature-check cases with mocked codesign output).
- `lint`, `typecheck`, `format:check`, and `build` all green.
- Fresh bare clone of this branch (no install): initialize handshake returns 2.5.8 and `tools/list` returns all 34 tools.
- Version bumped to 2.5.8 with `scripts/sync-plugin-version.mjs` run, per version-guard.

## Deliberately left out

- No IMAP-style connection changes or new tools; this is policy and diagnostics only.
- I considered making the retry default 3 (the value the old doc comment suggested for opt-in) but chose 2: timeouts are classified transient, and with a 30s timeout each attempt is expensive, so one retry is the sane default and the env knob covers anyone who wants more.


sweetrb:
Landed in **v2.5.8** via #71. Your branch was stacked on #69, so once #69 merged it needed re-basing onto main — I re-applied your reliability changes cleanly and regenerated the bundle, crediting you in the CHANGELOG. Thanks for the trio, @oliverames: the transient-only retry default (1→2), the `envPositiveNumber` env knobs following the existing `MAX_BUFFER` validation pattern, and especially the `checkNodeRuntimeSignature` doctor check — an ad-hoc-signed Node silently losing its TCC grants on every update is exactly the failure mode we've fought. Closing in favor of #71.

PR #72, opened 2026-07-06
Fix list-attachments always returning empty: reserved word `item` broke script compilation

oliverames:
## What happened

Both `listAttachmentsById` and `listAttachments` build their return value with a second loop, `repeat with item in attachmentList`. AppleScript refuses to compile that line because `item` is a class name ("Expected variable name or property but found class name", error -2741). Every list-attachments call has been failing before any Apple Event reached Notes, and the error handler turns that into an empty array.

The silent empty array is the part that worried me. The tool descriptions tell callers to run list-attachments as a safety check before `update-note`, which replaces the whole note body and drops attachments. A false zero waves that check through. I hit this on a real note with 7 images; the tool said 0 while `count of attachments` via osascript said 7.

## The fix

- Rename the loop variable to `recordItem` in both methods.
- Normalize the literal string `"missing value"` (what `URL of a as text` yields when an attachment has no URL, which is most images) to `undefined` in the parsed `url` field, so the sentinel doesn't leak into results.

## Tests

The unit tests mock `executeAppleScript`, so a compile-time failure in the generated script is invisible to them. I added a regression test that captures the generated script and asserts the reserved word isn't used as a loop variable, plus a parser test for the `"missing value"` URL case. `lint`, `format:check`, `test` (450), and `build` all pass.

## Verified live

macOS 27.0, Notes 4.13. Against the 7-attachment note: count 0 before the fix, 7 after, and `url` is absent instead of `"missing value"`. JSON-RPC probe against the built bundle confirms the same through the MCP surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.5.9] - 2026-07-06
+### Fixed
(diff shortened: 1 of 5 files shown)


oliverames:
Closing this in favor of #75, which bundles this fix with the related save-attachment and get-note-content media fixes. That should make review easier and keeps the release bump to one 2.5.9 entry.

PR #73, opened 2026-07-06
Fix save-attachment: successful saves misreported as failures, plus path and error-clarity hardening

oliverames:
## The headline bug

`saveAttachmentById` builds its return value as `return "OK${AS_FIELD_SEP}" & ...`. The interpolation lands inside the AppleScript string literal, so the script returns the literal text `OK(ASCII character 31)...` rather than a real control character. The TypeScript side splits on `\x1f`, finds nothing, and reports "attachment not found" even though the file was written to disk. As far as I can tell every successful save-attachment call has been misreported this way; the unit tests never caught it because they mock `executeAppleScript` and hand the parser correctly separated strings.

I found this the honest way: a session where `ls` showed the saved file sitting on disk while the tool insisted the attachment didn't exist.

## Also in this change

- `allowedSaveRoots` accepted `/tmp` but rejected `/private/tmp`, which is the real directory behind macOS's `/tmp` symlink. Same rationale as the existing `/private/var/folders` entry.
- Notes' `save` does not create intermediate directories and fails with an opaque "Failed saving an attachment to <path>" when the parent is missing. The manager now creates the parent directory before invoking Notes.
- Link-preview attachments (the rich preview Notes builds for a pasted URL) have no file payload, and `save` raises a bare "AppleEvent handler failed" for them. The error now explains that and includes the preview's URL so the caller has something actionable.
- `show-attachment` had the same quoted-literal quirk in its ERR return; it happened to work because it only checks `startsWith("ERR")`, but it is fixed for consistency.

## Tests

New regression test captures the generated script and asserts the sentinel protocol emits real separator expressions rather than text inside string literals, plus tests for parent-directory creation, the link-preview error, the plain error path, and the `/private/tmp` root. `lint`, `format:check`, `test` (455), and `build` all pass.

## Verified live

macOS 27.0, Notes 4.13. An image attachment saved into a not-yet-existing `/private/tmp` subdirectory returns success with metadata (previously: "attachment not found" despite the file landing), and saving an App Store link preview returns the explanatory error with its URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.5.10] - 2026-07-06
+### Fixed
+- **`save-attachment` misreported every successful save as "attachment not found".** The OK/ERR sentinel interpolated the field separator inside the AppleScript string literal (`return "OK${AS_FIELD_SEP}" & ...`), so the script returned the literal text `OK(ASCII character 31)...` instead of a control character and the TypeScript split never matched, even though the file landed on disk. Separators are now concatenated as expressions, matching the list methods; `show-attachment` had the same quirk in its ERR return and is fixed for consistency. A regression test inspects the generated script for separators inside string literals. (#73)
+- **`/private/tmp` destinations are allowed.** `allowedSaveRoots` accepted `/tmp` but rejected `/private/tmp`, the real directory behind macOS's `/tmp` symlink. (#73)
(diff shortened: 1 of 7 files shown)


oliverames:
Version bump note: the three PRs I opened today carry 2.5.9 (#72), 2.5.10 (this one), and 2.5.11 (#74) on the assumption they merge in that order. If you take them in a different order or only some of them, say the word and I'll rebump.

oliverames:
Closing this in favor of #75, which bundles this fix with the related list-attachments and get-note-content media fixes. That should make review easier and keeps the release bump to one 2.5.9 entry.

Issue #16, opened 2026-06-19
execSync has no maxBuffer cap (ENOBUFS on large output)


sweetrb:
src/utils/applescript.ts:324 and src/utils/jxa.ts:103 call execSync with no maxBuffer, inheriting Nodes 1MB default. export-notes-json, getRecentlyModifiedCounts, get-note-content on long notes can exceed it → execSync throws and the catch returns []/null (silent, data-dependent failure). Fix: add a 64MB default maxBuffer (env-overridable) to both executors, mirroring apple-mail #27.

sweetrb:
Implemented in the 2.0 maturity pass. Both executors set a 64 MB `maxBuffer`, env-overridable via `APPLE_NOTES_MCP_MAX_BUFFER`: `getMaxBuffer()` in src/utils/applescript.ts (~L28, applied ~L373) and src/utils/jxa.ts (~L24, applied ~L122). No more silent ENOBUFS on export-notes-json / long get-note-content. Closing as done.

PR #74, opened 2026-07-06
Cap inline base64 images in get-note-content so image-heavy notes can't kill the connection

oliverames:
## What happens today

Notes returns a note's `body` HTML with pasted images embedded as base64 `data:` URIs. A note holding a handful of photos easily produces a 10+ MB body, and a tool response that large can exceed the MCP client's message limit and take down the whole stdio transport. I have a real note that reproduces this: 7 pasted images, 10.5 MB body, and every `get-note-content` call on it ends in "MCP error -32000: Connection closed", after which the server is gone for the rest of the session. A 5 MB note on the same machine goes through fine, so the ceiling sits somewhere between the two. The 64 MB osascript buffer from #16 protects the AppleScript hop, but nothing bounded what we hand back to the client.

## The fix

`get-note-content` now runs the body through a small utility that replaces each inline image whose base64 payload exceeds a per-image cap with a plain placeholder naming the media type and approximate decoded size. When anything was stripped, the response text gets a warning that says how many images were omitted and points at `list-attachments` plus `save-attachment` / `fetch-attachment` for exporting the real files, which is the better path for image payloads anyway.

The default cap is 256 KB per image, so small pasted images (icons, sketches, modest screenshots) stay inline and photos get placeholders. `APPLE_NOTES_MCP_MAX_INLINE_IMAGE_BYTES` overrides it, following the same env-knob pattern as the timeout and retry settings. Note text is never touched.

One scope choice worth naming: this caps per-image rather than per-response, so a note with dozens of small images could in theory still build a large response. I kept it simple because the failure I can actually reproduce is photo-sized images, and a per-response cap would have to decide which content to drop.

## Tests

Unit tests cover the cap default and env override, under- and over-cap images, multiple images, single-quoted attributes, non-data `img` tags, and the warning text. `lint`, `format:check`, `test` (458), and `build` all pass.

## Verified live

macOS 27.0, Notes 4.13. The 10.5 MB note that previously killed the connection now returns a 19 KB response with all note text intact, 7 placeholders where the photos were, and the warning appended.


oliverames:
Same version-bump note as #73: this one carries 2.5.11 assuming #72 and #73 land first. Happy to rebump if the order changes.

oliverames:
Closing this in favor of #75, which bundles this fix with the related list-attachments and save-attachment fixes. That should make review easier and keeps the release bump to one 2.5.9 entry.

PR #75, opened 2026-07-08
Bundle Notes media reliability fixes

oliverames:
## Description

This bundles #72, #73, and #74 into one patch release. The three fixes all came from the same attachment-heavy failure path: callers need to know what attachments exist, save them reliably, and avoid crashing the MCP connection when `get-note-content` returns pasted images inline.

The branch keeps the three behavior commits intact, then adds one 2.5.9 release/docs commit so the version history reads as one bundled release instead of three separate patch bumps.

It fixes:

- `list-attachments` returning an empty list because the generated AppleScript used `item` as a loop variable.
- `save-attachment` writing files successfully but reporting "attachment not found," plus `/private/tmp`, missing parent directory, and link-preview error handling.
- `get-note-content` returning photo-sized inline `data:` URIs large enough to drop the MCP stdio connection.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Testing

- [x] Tests pass locally (`npm test`)
- [x] Linting passes (`npm run lint`)
- [x] Build succeeds (`npm run build`)
- [x] `pnpm run format:check`
- [x] `pnpm run typecheck`
- [x] `pnpm run test:coverage` (17 files, 467 tests)
- [x] `pnpm run test:integration` (2 files, 14 tests, including live Notes.app create/read/search/delete coverage)

## Checklist

- [x] I have updated documentation if needed


sweetrb:
Merged and released as **v2.5.9** — thanks for bundling #72/#73/#74 into one clean patch, @oliverames. Three genuinely nasty ones:
- the `item` reserved-class loop variable compiling to a silent `-2741` and returning **zero** attachments — and catching the second-order impact (it defeats the attachment-safety check callers run before `update-note` replaces the whole body) is the sharp part;
- the field separator interpolated *inside* the AppleScript string literal, so `save-attachment` reported "attachment not found" on saves that actually landed on disk;
- the inline `data:` URI cap so an image-heavy note can't exceed the client's MCP message limit and drop the stdio transport.

The regression tests that inspect the **generated script** for reserved loop vars / separators-inside-literals are a great touch — exactly the class of bug the mocked unit tests can't see. Verified the 2.5.9 bundle runs standalone with no `node_modules`.

Issue #17, opened 2026-06-19
No with-timeout wrap + SIGTERM kill can wedge Notes.app


sweetrb:
applescript.ts wraps nothing in `with timeout` and relies on execSync default SIGTERM. Killing osascript with SIGTERM leaves work dispatched inside the single-threaded app running, wedging later calls (cf apple-mail #11). Fix: port wrapWithTimeout (script-level timeout below process timeout) + killSignal:SIGKILL.

sweetrb:
Implemented. `wrapWithTimeout()` (src/utils/applescript.ts:53) wraps every script in an AppleScript `with timeout` block with 5s headroom below the process timeout, and both executors use `killSignal: SIGKILL` (applescript.ts:370, jxa.ts:121) so a wedged osascript is reaped reliably instead of leaking work into Notes.app. Closing as done.

PR #79, opened 2026-07-08
Run osascript without a shell, and fix timeout detection that never fired in production

oliverames:
## What this fixes

While reviewing the execution layer I found two related problems, both in how `executeAppleScript` and `executeJXA` shell out.

**Timeouts were never detected in production.** `isTimeoutError` checks `killed === true || signal === "SIGTERM"`, which is the error shape of the async `exec` API. A timed-out `execSync` call actually throws the underlying spawnSync error: `code: "ETIMEDOUT"` with the configured kill signal (`SIGKILL`, from #17). So a real timeout fell through to generic error parsing and surfaced as the raw `spawnSync /bin/sh ETIMEDOUT`. Worse, the retry gate keys off that detection, and `ETIMEDOUT` does not match the `/timed? out/i` transient pattern, so the retry-on-timeout behavior from #70 never actually ran. The unit tests passed because their mocked errors used the async shape. Detection now checks `ETIMEDOUT`/`SIGKILL` first (the old checks stay as a fallback for anyone wrapping this with async exec), and the tests throw the real error shape.

**Both executors composed `osascript -e '<script>'` as a shell string.** That made single-quote escaping the only barrier between note content and arbitrary shell execution, and it capped generated script size at the kernel argv limit, so a large enough note body could fail with E2BIG. They now call `execFileSync("osascript", ["-"], ...)` with the script delivered over stdin. No shell is involved at all, so the injection class of bug is structurally impossible, script size is unbounded, and each call saves a `/bin/sh` fork. `escapeForShell` and the JXA equivalent are gone. As a small bonus the retry sleep now blocks in-process via `Atomics.wait` instead of forking `/bin/sleep` per attempt.

## The honest caveat

The timeout bug survived this long because the tests encoded the wrong error shape, so the mocks agreed with the bug. I rewrote the affected tests to use the shape Node actually throws, and I verified against the real thing before trusting them: probed Node's sync-exec timeout error directly, confirmed `osascript -` reads both AppleScript and JXA from stdin, then ran a live check against Notes.app. A real `count of notes` call succeeds over stdin, and a forced 3-second timeout now returns the friendly timeout message instead of the raw spawnSync error.

## Out of scope

I did not touch the AppleScript-literal escaping in `appleNotesManager` (that layer is correct and unrelated), and I left the `with timeout` wrapping and SIGKILL semantics from #17 exactly as they were. No behavior changes for callers beyond timeouts now being reported and retried as #70 intended.

`lint`, `typecheck`, `test` (468 passing), and `build` are all green.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.8",
+      "version": "2.5.10",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.8",
+      "version": "2.5.10",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-notes",
-  "version": "2.5.8",
+  "version": "2.5.10",
   "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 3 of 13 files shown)


sweetrb:
Merged — thank you, @oliverames. Two things I especially appreciated here: catching that `isTimeoutError` was matching the *async* exec shape (`killed`/`SIGTERM`) while `execFileSync` actually throws `code: ETIMEDOUT` with the SIGKILL kill signal — so timeouts silently fell through *and* the #70 retry gate never fired — and the honest note that the mocked tests had been agreeing with the bug (rewriting them to Node's real error shape is exactly right). Moving both executors to `execFileSync("osascript", ["-"], { input })` over stdin is a clean structural kill of the shell-injection surface and the ARG_MAX cap, and the `Atomics.wait` sleep is a nice bonus. Verified the JXA path keeps `-l JavaScript` and the timeout detection matches the real error shape before merging.

PR #81, opened 2026-07-14
Fix stale sync warnings from orphaned cloud-state rows

oliverames:
## What changed

`get-sync-status` was reporting 158 pending uploads on my Mac even though Notes was healthy and the count never changed. I traced those rows through `NoteStore.sqlite`. All 158 were orphaned `ZICCLOUDSTATE` rows, and none were referenced by a live `ZICCLOUDSYNCINGOBJECT`.

This change requires that live reference before a version gap counts as pending. Real gaps on live objects still count; historical orphan rows do not.

## Verification

- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm test` (469 tests)
- `pnpm run build`
- Live MCP handshake and `get-sync-status`: 34 tools, `pendingUpload: 0`, idle

I kept this scoped to sync detection and its regression. It does not change Notes data or write behavior.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.11",
+      "version": "2.5.12",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.11",
+      "version": "2.5.12",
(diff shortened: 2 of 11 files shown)


sweetrb:
Merged — thanks @oliverames. What made this one easy to take: you traced the actual 158 rows through NoteStore.sqlite to the orphaned `ZICCLOUDSTATE` root cause rather than papering over the symptom, and the `EXISTS` guard against a live `ZICCLOUDSYNCINGOBJECT` is exactly the right shape — real version gaps on live objects still count, historical residue doesn't. Also appreciated the complete release packaging (2.5.12 bump + CHANGELOG + manifests + rebuilt bundle riding the PR — you clearly know the house rules by now) and the regression test pinning the join. One note from my side: I verified the change is fail-safe on machines where the sqlite read is TCC-denied (degrades to the existing error path, no false positive), so the worst case anywhere is no worse than before. 2.5.12 will be on npm shortly.

PR #82, opened 2026-07-15
Add append-to-note tool

vijaYE-bass:
## Summary

Adds two new tools designed to work together for a synthesis workflow — batching related notes from a task manager into Apple Notes documents.

### append-to-note
Adds content to an existing note without replacing the full body. Callers pass `id` (preferred) or `title` plus the content to add — no need to manually orchestrate `get-note-content` then `update-note`.

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `id` | string | — | Note ID (preferred) |
| `title` | string | — | Note title (fallback) |
| `content` | string | — | Text to append |
| `position` | `"after"` or `"before"` | `"after"` | Append to end or prepend to start |
| `separator` | string | `"\n\n"` | Separator between existing and new content |
| `format` | `"plaintext"` or `"html"` | `"plaintext"` | Format of the appended content |
| `account` | string | — | Account (ignored if `id` provided) |

### get-note-link
Returns the `notes://showNote?identifier=<uuid>` deep-link URL for a note. This is the same URL Notes.app copies via Copy Note Link — tapping it on iOS or macOS opens the note directly. Designed to be used after creating or appending to a note so the URL can be stored in a linked task or reminder.

| Parameter | Type | Description |
|-----------|------|-------------|
| `id` | string | Note ID (preferred) |
| `title` | string | Note title (fallback) |
| `account` | string | Account (ignored if `id` provided) |

Requires macOS 12+; returns a descriptive error on older systems.

## Motivation

The common append pattern requires two MCP calls and error-prone concatenation on the caller side. `append-to-note` wraps the read-concatenate-write cycle in a single call. `get-note-link` completes the loop by making the resulting note linkable from external tools without the caller needing to know the notes:// URL format.

## Test plan

- [ ] All existing tests pass (`npm test` -- 469 tests)
- [ ] Append to a note by ID, verify content added at end
- [ ] Prepend to a note using `position: "before"`
- [ ] Custom separator renders correctly
- [ ] Error returned for unknown note ID
- [ ] Error returned for password-protected note
- [ ] get-note-link returns a valid notes:// URL by ID
- [ ] get-note-link returns a valid notes:// URL by title
- [ ] Error returned for unknown note ID in get-note-link

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.12",
+      "version": "2.6.0",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.5.12",
+      "version": "2.6.0",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-notes",
-  "version": "2.5.12",
+  "version": "2.6.0",
   "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,7 +11,7 @@
     {
       "name": "apple-notes",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, and organize notes and folders",
-      "version": "2.5.12",
+      "version": "2.6.0",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-notes",
-  "version": "2.5.12",
+  "version": "2.6.0",
   "description": "Manage Apple Notes through natural language - create, search, read, update, delete, and organize notes and folders (macOS only)",
   "author": {
     "name": "Rob Sweet",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.6.0] - 2026-07-16
+### Added
+- **`append-to-note`**: Appends or prepends content to an existing note by id or title, preserving all existing rich HTML formatting (bold, italic, etc.). Always reads and writes as HTML, splitting the title `<div>` from body to prevent title duplication. Supports `position` (`"after"` / `"before"`), `separator`, and `format` (`"plaintext"` / `"html"`) parameters.
+- **`get-note-link`**: Returns the `notes://showNote?identifier=<uuid>` deep-link URL for a note by id or title. Primary path queries the Notes SQLite database for `ZIDENTIFIER` (works on all macOS versions including macOS 26+); falls back to the AppleScript `note link` property on macOS 12–15. Skips password-protected notes.
+
 ## [2.5.12] - 2026-07-13
 ### Fixed
 - **`get-sync-status` no longer reports orphaned Core Data rows as pending uploads.** The detector counted every `ZICCLOUDSTATE` version gap, including historical rows whose Notes syncing object no longer exists. It now requires a matching live `ZICCLOUDSYNCINGOBJECT` reference before treating a row as pending, preventing a permanent false active-sync warning while preserving detection for live unsynced objects.
--- a/README.md
+++ b/README.md
@@ -469,6 +469,71 @@ Moves a note to a different folder. The note is relocated in place via Notes.app
 
 ---
 
+#### `append-to-note`
+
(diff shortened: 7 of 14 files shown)


sweetrb (changes requested):
Thanks @vijaYE-bass — both tools are genuinely wanted here. `get-note-link` especially: I verified your macOS-26 claim on a real 26.5 machine (`sdef Notes.app` — the `note link` property is indeed gone), so the SQLite approach is the right forward path, and your CoreData `p<PK>` → `ZIDENTIFIER` mapping is correct. I also rebuilt `build/index.js` from your source — byte-identical to your committed bundle — and all repo gates pass (lint/format/typecheck/469 tests). Good structured tool descriptions too.

That said, I live-tested against a real Notes library and found issues that need fixing before merge:

**1. Default `append-to-note` corrupts rich notes (data loss — must fix).** With `format: "plaintext"` (the default), the read-concatenate-write round-trip destroys the existing note: I created a note with `<b>`/`<i>` content, appended one plain line with defaults, and the result had **all formatting flattened to plain `<div>`s and the note title duplicated into the body** (the plaintext read includes the title line; the write re-embeds it). Suggested fix: always read the existing body as **HTML** (`getNoteContentById`), strip the leading title `<div>` the way `update-note` handles it, and append the new content as an HTML-escaped `<div>` block when `format:"plaintext"` — so the appended text is plain but the existing note is untouched. Please add a regression test that asserts `<b>` survives a default append and the title appears exactly once.

**2. `new DatabaseSync(dbPath)` opens Apple's live NoteStore.sqlite read-write** (node:sqlite's default). Please pass `{ readOnly: true }` — we never want a write handle (or lock state) on a database Notes.app owns. (Our own `syncDetection.ts` uses `sqlite3 -readonly` for the same reason.)

**3. Misleading error remediation.** "requires macOS 12 or later" is wrong on macOS 26: there the SQLite path is the *only* path, and its usual failure cause is missing **Full Disk Access** (the AppleScript fallback no longer exists). Please use the house pattern — the FDA remediation string + absolute guide URL from `src/utils/docsUrls.ts` ("grant Full Disk Access to the app that launches the server, fully quit and relaunch, run the `doctor` tool") — and mention the macOS-12+ constraint only for the legacy fallback.

**4. Tests.** Two new tools with zero coverage in a 469-test repo. Minimum: the flatten/title regression above, append by-id happy path + position/separator, `get-note-link` PK-parse + identifier-to-URL mapping (the pure functions are easy to cover), password-protected rejection.

**5. Release packaging (CI `require-version-bump` will fail as-is):** `pnpm version minor --no-git-tag-version` (new tools = minor), `node scripts/sync-plugin-version.mjs`, a CHANGELOG entry under the new version (`### Added`, one bullet per tool), rebuild + commit the bundle. Also add both tools to the README Tool Reference (see the existing table style).

Minor (no action required): `get-note-link` by-id fetches the note twice (handler `getNoteById` + `getNoteLinkById`'s internal re-fetch) — AppleScript round-trips are the expensive part here, worth collapsing while you're in there.

Push the fixes to this branch and I'll re-review same-day — the underlying design is right, and both tools will ship.

vijaYE-bass:
Thanks for the thorough review — here's a second attempt:

**1. `append-to-note` HTML corruption (data loss fix)**
Both paths now always read existing body as HTML, split the title `<div>` using `indexOf("</div>")`, convert plaintext content to `<div>`-wrapped HTML, and write back as HTML. Bold/italic survives a default append; title appears exactly once. Regression test added: `<b>` tag survives append, title does not duplicate.

**2. `DatabaseSync` read-write handle**
Changed to `new DatabaseSync(dbPath, { readOnly: true })` — same pattern as `syncDetection.ts`.

**3. Error message**
Replaced "requires macOS 12 or later" with FDA remediation text + `FULL_DISK_ACCESS_GUIDE_URL`. macOS 12–15 fallback note preserved.

**4. Tests**
8 new tests across `appleNotesManager.test.ts` and `index.test.ts`: `getNoteLinkById` happy path + password-protected rejection, append-after/before position, bold preservation, title-not-duplicated assertion.

**5. Release packaging**
Bumped to `2.6.0`, `sync-plugin-version.mjs` run, CHANGELOG `### Added` entry for both tools, README Tool Reference sections added.

(Double-fetch in `getNoteLinkById` left as-is — the two calls are different data sources and collapsing them would be a non-trivial refactor.)

_One gap: `getNoteLinkFromDB` is module-private so the CoreData PK-parse (`/p(\d+)$/` → Z_PK) isn't directly covered — happy to export it and add a pure-function test if you'd prefer that over the current indirect coverage._

sweetrb (approved):
Verified — all five items are addressed:

1. **HTML corruption fix** — confirmed via the new regression tests (`<b>` survives a default append, title `<div>` appears exactly once) and by reading the updated logic: both id- and title-based paths now always read via `getNoteContentById`/`getNoteContent` (HTML), split the title `<div>`, convert plaintext to escaped `<div>` blocks, and write back as `"html"`.
2. **`DatabaseSync(dbPath, { readOnly: true })`** — confirmed, matches the `syncDetection.ts` pattern.
3. **Error message** — now uses the `FULL_DISK_ACCESS_GUIDE_URL` remediation string, with the macOS 12–15 fallback caveat preserved correctly.
4. **Tests** — 8 new tests added (append-after/before, bold preservation, title-not-duplicated, `getNoteLinkById` happy path + password-protected rejection + PK-parse sanity). Ran the full suite locally: 477/477 pass, plus clean typecheck/lint/format.
5. **Release packaging** — `2.6.0` bump across `package.json` + all plugin manifests (ran `sync-plugin-version.mjs`), CHANGELOG `[2.6.0]` entry, README Tool Reference sections for both tools. Rebuilt `build/index.js` from your source — byte-identical to your committed bundle.

The PK-parse gap you flagged for `getNoteLinkFromDB` is fine left as indirect coverage — no need to export it just for that. Merging.

sweetrb:
Merged — thanks for the thorough second pass, @vijaYE-bass. The HTML-preservation fix in particular was handled exactly right (title-div split + always-HTML round trip), and the regression tests you added for it (bold survives, title not duplicated) are exactly what I'd have asked for. Nice find on the macOS 26 `note link` SDEF removal too — that's a real gap the SQLite fallback closes well. `2.6.0` will publish to npm shortly.

PR #86, opened 2026-07-20
perf: bulk-fetch note properties in listNotes instead of per-note Apple Events

oliverames:
listNotes iterates notes with a repeat loop that sends two Apple Events per note (name of n, id of n). That works fine on small libraries, but it scales at roughly 8 notes per second. On my 524-note library a full listing takes 63 seconds, which blows past the 60 second tool timeout that MCP clients like Claude enforce. Since health-check runs an unbounded listNotes internally, a large library makes the connector look completely broken to the client even though nothing is wrong.

This change fetches name of notes and id of notes as two whole-list Apple Events and zips them in a local AppleScript loop. The modifiedSince filter needed one more step: a whose clause gets evaluated per note on the Notes side, so it is just as slow as the explicit loop. Instead I bulk-fetch modification date of notes as a third list and compare against thresholdDate locally. Comparing AppleScript dates in AppleScript also sidesteps any locale issues from coercing dates to text and parsing them in JS, which is the one part I would flag for review.

Measured on the same 524-note library through an MCP client: full listing with modifiedSince went from 63s (timeout) to 6.7s, and health-check went from 60s+ (timeout) to about 10s.

Dedup and the limit now happen in JS after the bulk fetch. I left the whose clause in searchNotes alone since it filters on note content, which has no bulk equivalent, and search result sets are small anyway.

Checks: npm run lint, typecheck, format:check, and npm test (477 passing, including three listNotes tests updated for the new script shape).

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.0",
+      "version": "2.6.1",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.0",
+      "version": "2.6.1",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-notes",
-  "version": "2.6.0",
+  "version": "2.6.1",
   "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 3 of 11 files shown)


sweetrb:
Thank you for this, @oliverames — the diagnosis is exactly right, the measurements made the case airtight, and the `whose`-clause insight (that Notes evaluates it per-note, so it's no better than the explicit loop) is the kind of thing you only learn by actually profiling. This was also the second time you'd found `health-check`'s unbounded listing at the bottom of a real-world failure, and the fix was overdue.

I've landed your work in #87 with your three commits cherry-picked as-is (authorship preserved), plus a hardening layer on top before release:

- **Mid-listing mutation guard** — the three whole-list fetches are separate Apple Events over a live, syncing collection, zipped by index. A library mutation landing between them silently mispairs names/ids (grow) or hard-errors unretried (shrink); I reproduced both in `osascript`. Each script now guards the list lengths match and raises a retryable `"Notes changed during listing"` error so the whole listing re-runs on a fresh snapshot.
- **`limit` slicing** — the old loop could `exit repeat` at the limit, so bounded calls were O(limit); the bulk fetch made them O(library). Bounded unfiltered listings now fetch `notes 1 thru N` with a total-count header and fall back to a full fetch on dedup shortfall, so your dedup-before-limit semantics are preserved exactly.
- **Your review flag on the date handling was warranted, in a way you didn't have line of sight to**: `buildAppleScriptDateVar` (pre-existing) assigns year → month → day onto `current date`, which rolls over when run on a day-of-month the target month lacks — June 15 requested on the 31st lands on July 15. Your PR made that helper the sole mechanism behind `modifiedSince`, so it's fixed in #87 (day pinned to 1 first).

Ships as 2.6.1 momentarily. Closing this in favor of #87 — your commits and the CHANGELOG credit you. Much appreciated, as always.

PR #93, opened 2026-07-20
fix: keep retries within deadline and run mutations once

oliverames:
## What changed

- Treats `timeoutMs` as one deadline shared by the initial AppleScript attempt, retry delays, and any retry attempts.
- Runs writes and other side-effecting AppleScripts once, even when read-only operations still allow transient retries.
- Adds regression tests, updates the configuration docs and changelog, and rebuilds the committed bundle.

## Why

I hit this while creating an Apple Note. A read call stayed active after the MCP client stopped waiting, and two create calls that appeared to time out completed later. That left three copies of the note.

The executor applied the default 30-second timeout to each of two attempts, with a one-second delay between them. That made the worst case about 61 seconds, just beyond the 60-second limit used by the client. Mutating calls inherited the same retry policy even though a timeout does not prove the write failed. Notes.app may have applied the change before `osascript` lost the response.

This keeps retries useful for reads without letting them extend the configured operation deadline. It also avoids replaying a write when the first result is unknown.

## Verification

- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm test` (487 tests)
- `pnpm run build`
- Live non-destructive probe: a five-second AppleScript delay with a 1.2-second total budget returned after 1.203 seconds and did not start a second full-budget attempt.

I left async request cancellation out of scope. The executor is still synchronous, but default calls now finish within their configured budget and mutations are not replayed.


sweetrb:
Thanks for all four of these — this was a genuinely good batch, and I've merged three of them tonight as 2.6.3, 2.6.4 and 2.6.5.

What landed:
- **#94** → 2.6.3 — `search-notes` real timestamps
- **#96** → 2.6.4 — `update-note` reporting the visible HTML title (with #95's README line carried over)
- **#93** → 2.6.5 — the shared retry deadline and mutations-run-once

Each diagnosis was correct and independently verifiable in-tree, which is rarer than it sounds — the 61-second worst case in #93, the `new Date()` placeholders in `searchNotes`, and the `newTitle`-is-ignored-in-html mismatch all reproduced exactly as you described. The write-ups made review much faster.

I pushed some follow-up commits to your branches rather than sending them back to you, since they were mostly mechanical and I wanted these to ship tonight:

1. **Rebases and re-bumps.** All four PRs bumped to 2.6.3, and because those blobs were byte-identical, `package.json` *didn't* conflict — git happily auto-merged it, so the second PR through would have contributed no version bump at all. Serialised them to 2.6.3/2.6.4/2.6.5 with a bundle rebuild each (fork PRs don't get our automatic rebuild).
2. **#93 retry-headroom inversion.** `wrapWithTimeout` floors the in-script `with timeout` at one second, so once the shared budget dropped below that, a retry got e.g. a 90 ms process timeout wrapped in `with timeout of 1 seconds` — inverting the ordering the headroom exists for. The gate now requires a minimum attempt budget. Worth noting I first tried folding in the full 5 s headroom and it broke *your* deadline test, correctly — 1000 ms is the right constant because it matches the floor.
3. **#96 parser performance.** The script/style strip pattern had no end-of-input branch, so every unclosed block scanned to EOF and backtracked: 211 KB 63 ms → 844 KB 1039 ms, against a 5 MiB content ceiling. Now 0–2 ms, and unclosed blocks are consumed to EOF like a browser would.
4. Minor: an `on error` guard on #94's date reads (an unreadable date would otherwise drop the note *and* suppress its duplicate references, since the ID is recorded before the reads), the stale README max-retries row, and two transposed docblock labels in `src/types.ts`.

If any of that doesn't match what you intended, say so and I'll happily revisit. And if you're interested in more — the `resolveUpdateResponseTitle` signature really would read better as #95's named-options object, if you want to send that as a follow-up.

PR #95, opened 2026-07-20
fix: report visible title after HTML note updates

oliverames:
## What changed

`update-note` now derives its response title from the first rendered line of `newContent` in HTML mode. That matches the title Notes.app actually displays.

I found this while using the documented title-cleanup flow. The update succeeded and Notes showed `Sailing Helmet Recommendations`, but the tool response said `Note updated: " "` because it echoed `newTitle`, even though HTML mode ignores that field.

The parser handles inline tags, common HTML entities, and numeric entities. If the HTML has no rendered text, it keeps the current title. Plaintext update behavior is unchanged.

I did not add a post-write Notes.app read. That would add another Apple Event to every HTML update, and the submitted HTML already contains the title Notes uses.

## Testing

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm test` (490 tests passed)
- `pnpm run build`

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.2",
+      "version": "2.6.3",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.2",
+      "version": "2.6.3",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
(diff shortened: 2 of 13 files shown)


sweetrb:
Thanks for both of these, and for the unusually clear write-ups — the diagnosis was correct in each case: `updateNote`/`updateNoteById` discard `newTitle` in `format: "html"` mode, so the handler was echoing a title that was never applied.

#95 and #96 are the same fix, so I'm keeping **#96** and closing this one as a duplicate. That's purely on parser behavior, not on submission order — I tested both helpers directly against the same inputs:

- #96 strips `<script>`/`<style>` **contents** before parsing. Pasted HTML with a leading `<style>` block is common, and #95 reports the CSS as the note title (`"h1{color:red}Actual Title"`).
- #96 also collapses intra-line whitespace to match what Notes actually renders, guards lone surrogates, and handles `<br / >`.

For what it's worth, I think #95 had the better API shape — the named-options object reads more safely than #96's positional `(currentTitle, newTitle, format, newContent)`, where `format` sits between the two title-ish arguments. And your CHANGELOG entry was the correct one: #96 filed under `## [Unreleased]`, which our `publish.yml` can't scrape, so I've fixed that on the way in.

I'm also carrying your README `**Returns:**` line over to #96 — `update-note` was one of the few tools missing one, and your wording filled the gap well. So both halves of your work land; you just get one commit instead of two.

Really appreciate the contributions — this was a genuine bug and a well-targeted fix.

sweetrb:
Closing as a duplicate of #96 — see the comparison above. Your README line and CHANGELOG fix are being carried into #96.

PR #97, opened 2026-07-21
fix: bulk-fetch attachment metadata

oliverames:
## Summary

- fetch attachment properties as whole lists instead of issuing seven Apple Events per attachment
- zip the property lists locally and retry when a concurrent Notes mutation changes their lengths
- cover both ID-based and title-based attachment listing paths with a regression test

## Live reproduction

On a real note with 60 attachments, the released 2.6.5 server exceeded its default 30-second AppleScript timeout. The tool then surfaced an empty attachment list, which is unsafe because callers use `list-attachments` before whole-body note updates.

A direct bulk AppleScript property fetch completed in 0.68 seconds. With this patch, an ordinary MCP `list-attachments` call under the unchanged default timeout returned all 60 attachments in 8.35 seconds.

## Verification

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm test` (500 passed)
- `pnpm run build`
- live MCP handshake and `list-attachments` call against the 60-attachment note


sweetrb:
Reviewed and verified against a live Notes library — this is a good fix, and the diagnosis is exactly right. Merging once CI is green on the two commits I pushed to the branch.

**What I confirmed on real data.** Both rewritten call sites (id- and title-based) return byte-identical output to `main` on a 2-attachment and a 7-attachment note, and the speedup holds: 3.9s → 1.75s end-to-end on the 7-attachment note, which is ~2.4s → ~0.25s of actual AppleScript. Extrapolated, that matches your 60-attachment timeout report.

I also chased down what looked like a regression and it is **not** one, so recording it here so nobody re-opens it: dropping the `try` around `URL of a` reads like it would break URL-less attachments, but `missing value as text` coerces to the literal string rather than erroring, so the old `try` never fired for that case — and `normalizeAppleScriptText` maps the sentinel to `undefined` either way. Old and new are identical here.

**Two things I did add**, both consequences of moving from per-attachment reads to a zip:

1. **Identity guard.** `repeat with a in attachments of theNote` yields *by-ID* object specifiers (`«class atts» id "x-coredata://…"`), so the old loop re-resolved each attachment by stable identity and a record could never mix fields from two attachments. Zipping seven independently-fetched lists gives that up: a same-length reorder, or a delete plus an add, between Apple Events passes all six count guards and produces a record with one attachment’s id and another’s name and dates. That is not cosmetic, because `save-attachment`/`fetch-attachment` resolve bytes from that id — the wrong file would be written under the wrong name. The script now re-reads the ids after the other six fetches and compares them element-wise.

2. **No false empties.** This one is really your own bug report pointed one level up. `listAttachmentsById`/`listAttachments` returned `[]` on a hard AppleScript failure, and the tool layer renders `[]` as the cheerful `Note "X" has no attachments` — a *successful* response that is indistinguishable from a genuinely empty note. Since callers use this list to decide whether a full-body update is safe, that is precisely the hazard your CHANGELOG entry describes, and the guards added six new ways to reach it. Failures now throw; only a successful call with empty output reports an empty note.

The false-empty is reproducible on shipped code, which made it easy to confirm the fix:

```
budget=1200ms, note "Boat Info" (7 attachments)
  main:  Note "Boat Info" has no attachments      <- wrong, and reported as success
  fixed: Found 7 attachment(s) in "Boat Info"     <- the bulk fetch is fast enough

budget=700-900ms
  fixed: Error listing attachments: ... Operation timed out ...   <- never a false empty
```

Tests added for field alignment across a 3-attachment record set, the identity guard in both generated scripts, and failure-throws-rather-than-empty. The existing `returns empty array on error` test encoded the old swallow, so I rewrote it. 504 passing.

Two small notes, neither blocking: I softened the CHANGELOG’s "default 30-second AppleScript timeout" since `wrapWithTimeout` derives the in-script limit from the process budget, so the binding constraint is the operation budget rather than AppleScript’s default. And separately from this PR, `showAttachmentById` and `saveAttachmentById` still linear-scan with `id of a` per attachment, so resolving *one* attachment on the image-heavy notes you are targeting still costs up to N Apple Events — worth the same treatment in a follow-up if you are interested.

Thanks for this one. The live before/after measurement in the PR description, and specifically calling out that the empty list defeats the pre-update safety check, is what made the second issue findable — that framing is what sent me looking at the tool layer.

sweetrb:
Shipped in [v2.6.6](https://github.com/sweetrb/apple-notes-mcp/releases/tag/v2.6.6), now on npm.

One follow-up landed in #98 (v2.6.7) for the third item I mentioned above: `show-attachment` and `save-attachment` were still resolving their target with the same per-attachment scan you removed here. It turns out Notes accepts `attachment id "…" of theNote` directly, so it is one Apple Event instead of up to N — and it still refuses to resolve an id belonging to a different note, which the scan was implicitly providing. Your fix is what made that worth looking for.

PR #99, opened 2026-07-21
fix: report the actual folder in search results

oliverames:
## What changed

`search-notes` was labeling every result as `Notes`, including notes in nested folders and Recently Deleted.

Notes.app throws when AppleScript reads `name of container of n` as one expression. Capturing the container first, then reading its name, returns the real folder. I added a regression test for the generated script and bumped the package and plugin versions to 2.6.8.

This keeps the existing response shape. Search results still report the containing folder's name rather than reconstructing its full nested path.

## Live verification

I reproduced this against my Notes library. With 2.6.7, an active note in `Personal/Tattoos` reported `Notes`. The patched build reports `Tattoos`, while two deleted duplicates report `Recently Deleted`.

## Checks

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm test` (507 tests)
- `pnpm run build`
- Live MCP call against nested and deleted notes

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.7",
+      "version": "2.6.8",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.7",
+      "version": "2.6.8",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-notes",
-  "version": "2.6.7",
+  "version": "2.6.8",
   "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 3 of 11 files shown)


sweetrb:
Merged. Correct diagnosis, minimal fix, and I confirmed every claim against a live library.

**The AppleScript failure reproduces exactly as you describe** — it is not a silent wrong answer, it is a hard error the `try` was swallowing:

```
CHAINED  (name of container of n): FAILED -1728:
    Notes got an error: Can't get name of container of note id "x-coredata://…/ICNote/p852"
TWO-STEP (container then name):    Notes
```

Live before/after on a note in a non-default folder:

```
2.6.7:  - Ichor Catch-Up — 2026-07-08 Pull (Notes)
2.6.8:  - Ichor Catch-Up — 2026-07-08 Pull (Ichor)
```

I also checked the fix is **complete**: line 914 was the only chained occurrence in the file. The three other `container of` reads (1629, 1976, 1978) already use the two-step form, which is presumably why nobody noticed this one.

**A bonus your description undersells.** `[DELETED]` at `src/index.ts:291` has always special-cased `folder === "Recently Deleted"`, but because folder was hardcoded to the `"Notes"` fallback in practice, that branch was **dead code**. Your fix revives it, so trashed notes now render as `- Search mcp-search-… [DELETED]` instead of `- … (Notes)`. That is a real safety improvement — an agent could previously have read a note out of Recently Deleted and treated it as live. Worth a CHANGELOG line of its own, in my view.

**One thing to be aware of, and it is not a blocker.** The old chained call failed fast; the new form issues two *successful* Apple Events per result. Measured over 100 notes on my library:

```
OLD chained + fallback:  5s
NEW two-step:            7s      (~+20ms per note)
```

That cost is inherent — resolving the container is what makes the answer right, so I would not trade it back. But it lands on `searchNotes`, which iterates **every** match when no `limit` is passed, and `limit` has no default. On a broad query against a 340-note library I measured 2.6.7 timing out at 30s on 2 of 3 runs, and 2.6.8 on 3 of 3. So the underlying fragility predates this PR — an unbounded search was already at the edge — but this pushes it over.

I am sending a follow-up that bounds it: a sensible default `limit`, plus memoizing the folder name per container inside the script (most results share a folder — my library resolves the same container 340 times). That keeps your correctness fix and removes the reason it hurts.

Thanks for this one, and for the trio today. The detail that made this reviewable in minutes rather than an afternoon was naming the exact failing expression and the exact folder you saw it in — that turns a vague "wrong folder" report into a one-line `osascript` repro.

PR #102, opened 2026-07-22
Skip npm publishing in forks

oliverames:
## What changed

This keeps the `Publish to npm` job scoped to `sweetrb/apple-notes-mcp`. Forks can keep the workflow file and run normal CI, but their publish job now skips before it requests an npm OIDC exchange or tries to create a GitHub Release.

I hit this in `oliverames/apple-notes-mcp`: once the fork's package version moved ahead of npm, its scheduled workflow tried to publish from the fork and failed the trusted-publisher exchange. The upstream triggers and release recovery behavior are unchanged.

## Checks

- `actionlint` 1.7.12 across all workflows
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run typecheck`
- `pnpm test` (507 tests)
- `pnpm run build`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,11 +45,12 @@ jobs:
   publish:
     runs-on: macos-latest
     if: >
-      (github.event_name == 'release') ||
-      (github.event_name == 'workflow_dispatch') ||
-      (github.event_name == 'schedule') ||
-      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success' &&
-        (github.event.workflow_run.event == 'push' || github.event.workflow_run.event == 'workflow_dispatch'))
+      github.repository == 'sweetrb/apple-notes-mcp' &&
+      ((github.event_name == 'release') ||
+       (github.event_name == 'workflow_dispatch') ||
+       (github.event_name == 'schedule') ||
+       (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success' &&
+         (github.event.workflow_run.event == 'push' || github.event.workflow_run.event == 'workflow_dispatch')))
     permissions:
       id-token: write
       contents: write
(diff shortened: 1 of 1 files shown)


sweetrb:
Thanks for this, Oliver — correct instinct, and a real gap: a fork that syncs this workflow fires `publish.yml` on push/schedule and either errors out (no OIDC trusted-publisher) or clutters the fork with red runs. Worth closing.

I pushed one change to your branch and rebased it onto current `main` (2.6.9). The change swaps `github.repository == 'sweetrb/apple-notes-mcp'` for **`github.repository_owner == 'sweetrb'`**, for two reasons:

1. **Conformance.** These four repos keep `publish.yml` **byte-identical** — `./conformance-check.sh` has it in the IDENTICAL set and fails on drift. Hardcoding one repo name makes notes' copy unique and can't be copied verbatim to the other three.
2. **Same result, portable.** `repository_owner == 'sweetrb'` is true on all four sweetrb repos and false on any fork, so it skips forks exactly as you intended while staying identical across the fleet.

The rebase was needed because #101 (search-notes limit) landed first and moved `main` to 2.6.9 — your branch was based on 2.6.8, so `require-version-bump` saw the stale base as an un-bumped diff. Rebased, the only change vs `main` is the one-line guard (workflow-only, version-exempt).

I'm landing the identical guard on the other three repos so all four stay in lockstep. Merging once CI is green.

sweetrb:
Landed as #103 (merged) — thank you, @oliverames! Credited via `Co-authored-by`.

Why a separate PR: a fork PR that modifies a workflow file can't be merged without the `workflow` OAuth scope, which the merge token here lacks. So I carried your commit forward on an in-repo branch (with a `repository == name` → `repository_owner == 'sweetrb'` tweak so it stays byte-identical across all four repos) and merged that. The same guard now lives in mail #106, numbers #38, and photos #46 — the whole fleet skips publish in forks. Nice catch.

PR #104, opened 2026-07-22
fix(security): override vulnerable MCP SDK transitive dependencies

oliverames:
## Description

This overrides two transitive dependencies that the MCP SDK currently cannot resolve to fully patched releases:

- `@hono/node-server` 2.0.10 clears the path traversal advisory affecting the 1.x release selected by the SDK.
- `fast-uri` 3.1.4 clears the two host confusion advisories affecting the version selected through AJV.

Dependabot cannot make either update on its own because both packages sit behind transitive version constraints. The exact overrides can come back out once the SDK widens its ranges. Hono 2 requires Node 20, which matches this package's existing runtime floor.

The shipped bundle is rebuilt, the package version is bumped to 2.6.10, and the changelog records the security fix.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other: dependency security fix

## Testing

- [x] `pnpm run lint`
- [x] `pnpm run format:check`
- [x] `pnpm run typecheck`
- [x] `pnpm test` (515 tests)
- [x] `pnpm run test:integration` (14 tests, including real server and Notes.app checks)
- [x] `pnpm run build`
- [x] `pnpm audit --audit-level low` (no known vulnerabilities)
- [x] Skill and plugin version sync checks

## Checklist

- [x] I have read the CONTRIBUTING guidelines
- [x] I have updated the changelog
- [x] The shipped bundle is committed and the package version is bumped

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.9",
+      "version": "2.6.10",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-notes",
-      "version": "2.6.9",
+      "version": "2.6.10",
       "description": "Manage Apple Notes through natural language - create, search, read, update, delete, organize folders, inspect metadata, run diagnostics, and work with attachments (macOS only).",
       "source": {
         "source": "local",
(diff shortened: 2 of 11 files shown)


sweetrb:
Merged as 2.6.10 — thank you, @oliverames. This is exactly the right fix: Dependabot couldn't touch either because they're transitive behind the SDK/AJV ranges, and a plain bump wouldn't stick. I verified the override resolves cleanly (fast-uri 3.1.4 in the shipped bundle, no 3.1.2 left) and, since fast-uri backs AJV's `format: uri` validation, smoke-tested a real MCP round-trip: valid input returns results and an out-of-range arg is still rejected with a clean -32602, so the bump doesn't break schema validation. I'm applying the same override to mail/numbers/photos now — they ship the same vulnerable fast-uri (3.1.2/3.1.3) — so the whole fleet clears it. Appreciate you jumping on the advisory.

PR #127, opened 2026-08-09
feat(list-notes): expose note ids alongside titles

### LINKED ISSUE #126: list-notes should expose note ids alongside titles
## Problem

`list-notes` returns `notes: string[]` — titles only, with no ids. A caller that enumerates notes via `list-notes` and then needs to act on a specific one (read, update, move, delete) has to re-resolve an id by title afterward, typically via `search-notes`.

That re-resolution step reintroduces the exact ambiguous-name problem `export-notes-json` was fixed for in #115: AppleScript's `note "<name>"` specifier resolves a duplicated title deterministically to one particular note every time. So two notes sharing a title, listed via `list-notes` and then looked up individually by title, silently collapse to the same note — one is fetched twice and the other is never reached, with no error or warning.

Concretely: a consumer syncing a folder's changed notes via `list-notes` + `modifiedSince`, then resolving each title to an id via `search-notes` for a follow-up read/update, will silently drop any note that shares a title with another in that result set.

The bulk AppleScript listing underneath `list-notes` already fetches ids for every note — `parseBulkListOutput` has returned `{title, id}` pairs since the #115 fix — so the id is available and simply isn't surfaced through `list-notes`' output.

## Proposed Solution

Add `AppleNotesManager.listNoteRefs(account?, folder?, modifiedSince?, limit?)` — same filtering as the existing `listNotes()` — returning `{title, id}` pairs instead of bare titles. `listNotes()` itself is unchanged and keeps returning `string[]` for its existing internal callers.

The `list-notes` MCP tool switches to calling `listNoteRefs()` and its response becomes `notes: Array<{title, id}>` instead of `notes: string[]`.

This is a public contract change to the `list-notes` tool response. It matches this project's own precedent for tool-schema changes shipped as a patch bump (e.g. the 2.6.9 `search-notes` limit-default change, also a breaking contract change) — version bump and CHANGELOG entry to follow the same pattern.

Implementation already drafted on branch `fix/list-notes-expose-ids` (rebased onto current `main`, build green, 523/523 tests passing):
- `src/services/appleNotesManager.ts` — new `listNoteRefs()`
- `src/index.ts` — `list-notes` tool now calls `listNoteRefs()`, updated `outputSchema` and description
- Tests for both, including a case with two same-titled notes distinguished by id

## Alternatives Considered

- **Change `listNotes()` in place** instead of adding `listNoteRefs()`. Rejected — `listNotes()` has existing internal callers that only need titles; changing its return shape would touch call sites with no benefit, whereas a parallel method keeps the internal `string[]` contract intact.
- **Leave `list-notes` as titles-only and document "re-resolve via `search-notes`"** as the workaround. Rejected — this is the trap that already bit `export-notes-json` (#115); documenting around it doesn't fix the silent-drop failure mode for duplicate titles.

## Additional Context

Related: #115 (`export-notes-json` duplicate-title fix), which this follows the same identity-safe pattern from (`getNoteById` / `getNoteContentById` over by-title lookups).

draku:
## Description

`list-notes` returns `notes: string[]` (titles only). A caller enumerating notes and then acting on a specific one has to re-resolve an id by title afterward — reintroducing the ambiguous-title trap `export-notes-json` was fixed for in #115: AppleScript's `note "<name>"` specifier resolves a duplicated title to one particular note every time, so two same-titled notes silently collapse into one when looked up individually.

Adds `AppleNotesManager.listNoteRefs(account?, folder?, modifiedSince?, limit?)` — same filtering as the existing `listNotes()` — returning `{title, id}` pairs pulled from the same bulk AppleScript listing that already fetches ids internally (`parseBulkListOutput`, since #115). `listNotes()` is unchanged and still returns `string[]` for its existing internal callers. The `list-notes` MCP tool now calls `listNoteRefs()`, and its response becomes `notes: Array<{title, id}>`.

This is a public contract change to the `list-notes` tool response — any caller destructuring `notes` as `string[]` will break. Shipped as a patch bump per this project's own precedent (e.g. the 2.6.9 `search-notes` limit-default change, also a breaking contract change patch-bumped rather than major-bumped).

Closes #126

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Testing

- [x] Tests pass locally (`pnpm test`) — 523/523
- [x] Build succeeds (`pnpm run build`)

## Checklist

- [x] I have updated documentation if needed
- [x] Version bumped (2.6.16 → 2.6.17) + CHANGELOG.md entry added

🤖 Generated with [Claude Code](https://claude.com/claude-code)


draku:
I'm not doing something quite right on the version bump. Will have to look more closely at the error that the check returns.

draku:
Fixed. I realize that this is a public contract change and could impact users coding against the api, rather than using an LLM tool call. Please let me know if you would recommend a different approach.

sweetrb:
Approach is right — merging this. Answering your question directly, since it was the right one to ask.

**On the contract change:** keep it. The alternative (leave `notes` as `string[]`, add a parallel `noteRefs` field) would preserve compatibility but leave the ambiguous-title trap as the *default* path, and your issue makes the case well that documenting around it is what already failed for `export-notes-json` in #115. The README, incidentally, has been claiming `list-notes` yields ids for a while — three separate `noteId` parameter tables say "from `search-notes`/`list-notes`". This makes the docs honest rather than breaking a promise.

**On the version, I bumped you 2.6.17 → 2.7.0.** The 2.6.9 precedent is real but weaker than it looks: that change altered a *default* while keeping the response type, so `notes.map(t => t.toUpperCase())` kept working and just saw fewer items. Yours changes the type, so that same line throws. CONTRIBUTING says "at least a patch" — a floor, not a ceiling — and a patch tells dependency tooling "safe, backwards-compatible," which this is not. Minor is the honest signal without a major-version churn on a server whose main consumer reads `outputSchema` at runtime anyway.

**One gap I fixed on your branch** (thanks for enabling maintainer edits): the ids reached `structuredContent` but not the human-readable text, which still rendered `  - <title>`. Plenty of MCP hosts surface only the text content to the model — for those, two same-titled notes still came back as two identical lines, so the exact failure mode #126 targets survived on that path. It now renders `  - <title> [id: <id>]`, matching what `search-notes` already does.

Also synced README (`**Returns:** List of note titles.`) and CLAUDE.md (`Returns note titles only`), both of which still described the old shape.

The refactor itself is the part I would call out as well done — routing both public methods through a private `listNotesCore()` instead of duplicating the fetch keeps the fast-path/full-fetch branch in one place, and the same-title test case pins the actual bug rather than just the happy path.

PR #128, opened 2026-08-11
fix: resolve Notes account by default-account/prefix match, not exact name

lakerfan0306:


lakerfan0306:
Closing — withdrawing this for now.

sweetrb:
Thanks for this, @lakerfan0306 — and sorry you withdrew it before I got to it. The diagnosis is the valuable part and I want it on the record rather than lost with the PR.

You identified something real that I had not considered: the implicit/default path hardcodes the literal `"iCloud"`, and a hardcoded account name is wrong for *any* reason the real name might differ — the U+F8FF suffix you hit, a localized account name, or a user whose default account simply is not iCloud. Replacing that with AppleScript's own `default account` is strictly more correct and I verified it here:

```
$ osascript -e 'tell application "Notes" to get name of default account'
iCloud
```

That half of your change I would take as-is.

The half I would not take as-is is the **explicit** account path — `tell (first account whose name starts with "X")`. Prefix matching plus `first … whose` is a silent wrong-target hazard: if the prefix matches more than one account, AppleScript picks whichever it lists first and reports success. On my own library:

```
$ osascript -e 'tell application "Notes" to get name of every account'
iCloud, rob@superiortech.io, robert.b.sweet@gmail.com

$ osascript -e 'tell application "Notes" to get name of (first account whose name starts with "rob")'
rob@superiortech.io
```

Two accounts start with `rob`; one is chosen, nothing surfaces the ambiguity. Applied across the ~10 call sites in the diff that includes `deleteNote` and the move paths, so the failure mode is a delete or move landing in the wrong account and reporting success. We fixed exactly this shape in the sibling repo recently (apple-mail-mcp #137, where `Archive` under two parents resolved to whichever the server listed first), and the resolution there was: exact match wins outright, a *unique* prefix match still resolves, and more than one match is refused with an error naming every candidate.

So the version I would ship is your `default account` fix as written, plus prefix matching for explicit names gated on uniqueness. That also closes the gap Copilot flagged — `moveNoteById` and `batchMoveNotes` still interpolate `of account "${safeAccount}"` directly, so they would have kept failing for the account name you were hitting even with this merged.

Two things would help me get it right, if you are willing:

1. What does `osascript -e 'tell application "Notes" to get name of every account' | xxd` show on your machine? I want the actual bytes of the account name — the suffix does not reproduce on macOS 26 here (mine is a clean `iCloud`, hex `69436c6f7564`), so I cannot regression-test against the real failure without knowing what it looks like.
2. Which tool call surfaced it for you, and what was the error text?

No obligation at all — if you would rather not, I will implement the hardened version from the analysis above and credit you for the diagnosis. Reopening this PR or opening a new one is equally welcome if you want to carry it yourself; the `default account` insight is yours either way.

sweetrb:
Shipped in #129, released as **2.7.1** — credited to you as `Co-authored-by` on the squash commit, since the diagnosis was the hard part and it was yours.

For the record, what your report got right and I had not considered: the implicit path hardcoded the literal `"iCloud"`, and that is wrong for *any* reason the real name might differ — your U+F8FF suffix, a localized name, or a default account that simply is not iCloud. It also could not be worked around from the caller's side, because the name you could see was not the name AppleScript was matching. That half went in as you wrote it, resolving through `default account`.

Two things I changed from your version, both about the *explicit* account path:

**1. Prefix matching is gated on uniqueness.** `first account whose name starts with "X"` picks whichever account AppleScript lists first and reports success. On my library:

```
$ osascript -e 'tell application "Notes" to get name of every account'
iCloud, rob@superiortech.io, robert.b.sweet@gmail.com
```

`"rob"` matches two. Across the ~10 account-scoped call sites in your diff that includes `deleteNote` and the move paths, so the failure mode is a delete or move landing in the wrong account, silently. It now resolves exact-match-first, unique-prefix-second, and *refuses* an ambiguous prefix naming every candidate:

```
Account "rob" is ambiguous - it matches 2 accounts:
rob@superiortech.io, robert.b.sweet@gmail.com. Use the full account name.
```

`account="robert"` still resolves to `robert.b.sweet@gmail.com`, so the ergonomics you were after survive.

**2. Closed the gap Copilot flagged on your PR.** `moveNoteById` and `batchMoveNotes` interpolate `of account "${safeAccount}"` directly and your diff didn't touch them — they would have kept failing for your account name even with it merged. Resolution now happens inside AppleScript and binds a single account reference, so those paths inherit it.

Also fixed while in there: an unresolvable account used to be reported as "note not found", which would have sent you looking for the wrong problem.

I never did get the byte-level detail of your account name, so there's no regression test against the exact U+F8FF case — if you do run `osascript -e 'tell application "Notes" to get name of every account' | xxd` sometime, I'd still like it. But 2.7.1 should work for you regardless now, since nothing matches on a hardcoded name any more. Please reopen or file a fresh issue if it doesn't.

Thanks for taking the time to diagnose it properly rather than just reporting "doesn't work" — and sorry the withdrawal nearly cost it.

PR #132, opened 2026-08-12
fix(schema): advertise JSON Schema 2020-12 so clients stop rejecting every tool

sweetrb:
## The problem

Current MCP clients (Claude Desktop among them) refuse **every tool** this server advertises:

```
Tool '<name>' has an invalid outputSchema: JSON Schema declares an unsupported
dialect ("$schema": "http://json-schema.org/draft-07/schema#"). The default
validator supports JSON Schema 2020-12 only.
```

This is not a degraded tool here and there — all **36 tools** become unusable at once. Reported against the sibling server as sweetrb/apple-mail-mcp#147; **all four Apple MCP servers are affected identically**, and this is the apple-notes-mcp half of that fix.

## Root cause

The dialect comes from the SDK, not from this repo:

- `@modelcontextprotocol/sdk`'s `server/mcp.js` calls `toJsonSchemaCompat(obj, { strictUnions, pipeStrategy })` with **no `target`**.
- In `sdk/dist/esm/server/zod-json-schema-compat.js`, `mapMiniTarget(undefined)` returns `'draft-7'`.
- The zod v3 branch then runs `zod-to-json-schema`, which defaults to draft-07 and stamps `"$schema": "http://json-schema.org/draft-07/schema#"` onto every emitted `inputSchema` **and** `outputSchema`.

MCP has since standardized on JSON Schema 2020-12 (SEP-834 / SEP-1613 / SEP-2106), so clients reject the older dialect outright.

### A zod v4 upgrade does NOT fix this

Checked before writing any code, and it does not help: with SDK 1.30.0 + zod 4.4.3 the emitted `$schema` is **still draft-07**. The SDK calls its converter without a target, so *both* its v3 (`zod-to-json-schema`) and v4 (`zod/v4-mini` `toJSONSchema`) branches fall back to `draft-7`. Bumping zod changes which branch runs, not what dialect it targets. No zod migration is attempted here.

## The fix

A small converter, `src/utils/jsonSchemaDialect.ts`, applied at the **transport boundary** — the only public seam that doesn't reach into SDK internals. `src/index.ts` changes one line:

```ts
const transport = withJsonSchema2020_12(new StdioServerTransport());
```

`withJsonSchema2020_12` wraps `transport.send` so every outgoing `tools/list` result has each tool's `inputSchema`/`outputSchema` re-declared as `https://json-schema.org/draft/2020-12/schema`. Every other message — tool-call results, notifications, errors — is returned by identity, untouched.

The converter also rewrites the keywords whose spelling or meaning changed in 2020-12:

| draft-07 | 2020-12 |
|---|---|
| `definitions` | `$defs` (and `#/definitions/X` refs → `#/$defs/X`) |
| tuple `items: [A, B]` | `prefixItems` |
| `additionalItems` (alongside a tuple) | `items` |
| `dependencies` | `dependentRequired` / `dependentSchemas` |
| draft-4 boolean `exclusiveMinimum`/`exclusiveMaximum` | numeric form, folding in `minimum`/`maximum` |

**That table is a no-op on today's schemas** — I verified the emitted bodies use none of those constructs, and the diff below confirms the schema bodies come out byte-identical apart from `$schema`. It exists so a future zod construct can't quietly reintroduce a dialect mismatch. (The one `$ref` in the family — apple-mail-mcp's `search-messages.properties.dateTo` → `#/properties/dateFrom` — is a relative pointer valid in both dialects and is left alone.)

**No tool definition, zod schema, or handler was changed.**

## Verification

**End-to-end MCP handshake against the rebuilt bundle** (`initialize` → `notifications/initialized` → `tools/list`, piped into `node build/index.js`), compared against the same probe run on `main`'s bundle:

| | main (2.7.1) | this branch (2.7.2) |
|---|---|---|
| tool count | 36 | **36** (names identical) |
| `draft-07` occurrences in the payload | 72 | **0** |
| `draft/2020-12` occurrences | 0 | **72** (36 input + 36 output) |
| inputSchemas not declaring 2020-12 | 36 | **none** |
| outputSchemas not declaring 2020-12 | 36 | **none** |
| nested `$schema` declarations | 0 | **0** (only the root declares a dialect) |
| `definitions` / `dependencies` / `additionalItems` | 0 | **0** |
| schema **bodies** ignoring `$schema` | — | **byte-identical to main** |

`additionalProperties: false` count is unchanged at 28 (input schemas only) — no regression against the #135/2.6.16 permissiveness fix.

**Tests**

- New `src/utils/jsonSchemaDialect.test.ts` — 26 tests covering the root stamp, nested-`$schema` stripping, `definitions`→`$defs` + `$ref` rewrite, tuple `items`→`prefixItems` with `additionalItems`→`items`, a stray `additionalItems` being *dropped* rather than clobbering a non-tuple `items`, the `dependencies` split, both boolean `exclusiveMinimum` collapses, non-mutation of the input message, pass-through of non-`tools/list` messages by identity, and that `withJsonSchema2020_12` returns the same instance, forwards `options`, and preserves the original `send`'s `this`. 100% statements/lines/functions, 98.18% branches on the new file.
- Extended the existing `test/output-schema.test.ts` contract suite (which boots the **real built server over stdio**) with a dialect assertion: every advertised schema must declare 2020-12 and carry no draft-07-only construct. Asserted against what the server actually advertises rather than the converter in isolation — the fix lives at the transport boundary, so a unit test alone could pass while the server shipped unwrapped. This test fails on `main`.

**Gate** — all green, no `--no-verify` anywhere:

```
pnpm run lint          ✅
pnpm run typecheck     ✅
pnpm test              ✅  557 passed (21 files)
pnpm run format:check  ✅
pnpm run test:coverage ✅  thresholds met
pnpm run test:integration ✅  11 passed / 5 skipped, contract suite 5/5
pnpm run build         ✅  committed bundle re-verified byte-identical
```

## Docs

Reviewed `README.md`, `CLAUDE.md` and all five files under `docs/`. The Tool Reference is untouched — no tool added, no parameter or output shape changed — and nothing documented the advertised schema dialect. Added one **Troubleshooting** entry to the README quoting the exact client error, since anyone hitting this sees every tool refused with no other symptom and nothing about their Notes library, permissions or config is involved.

## Release

Version bumped **2.7.1 → 2.7.2** with the CHANGELOG entry folded in, per the repo's mandatory-bump rule. `## [Unreleased]` held a `github-actions` pin note that this release necessarily drains (version-guard hard-fails a bump with a non-empty `[Unreleased]`), so it moved under `## [2.7.2] - 2026-08-12` and the marker is back to empty.


sweetrb:
### Follow-up: the converter is now position-aware (c626725)

Found while re-reading the converter added in this PR. `convertNode` recursed uniformly and then switched on **every key it met** — but the keys of a `properties` map are caller-chosen **tool parameter names**, not schema keywords. So the switch corrupted the tool contract for any tool that happened to name a parameter after a keyword:

| A tool parameter named… | …was rewritten to |
|---|---|
| `definitions` | renamed to `$defs` on the wire |
| `$schema` | **silently deleted** — while `required` still listed it, producing a schema **no input can satisfy** |
| `dependencies` | restructured into `dependentRequired` / `dependentSchemas` |
| `additionalItems` | dropped outright as a stray tuple partner |

Symmetrically, `enum` / `const` / `default` / `examples` hold instance **data**, not schemas, so recursing into them rewrote a caller's literal values — a `default` of `{"definitions": 1}` came back as `{"$defs": 1}`.

**The fix.** A key means a keyword only at a **schema position**. Two new module-level sets drive it: `SCHEMA_MAP_KEYWORDS` (`properties`, `patternProperties`, `$defs`, `dependentSchemas`) convert their **values** through a new `convertSchemaMap` and leave the caller's names untouched; `DATA_KEYWORDS` pass through verbatim. `definitions` keeps its own case — it *is* a keyword wherever it sits at a schema position — and the `dependencies` case still converts its schema values with `convertNode`, which was already correct.

Also documented the one deliberate limitation on the `$ref` case: only the exact root prefix `#/definitions/` is rewritten. A pointer *through* a nested `definitions` block (`#/properties/x/definitions/Y`) is left alone, because `#/properties/definitions/...` could equally address a property *named* `definitions` and the two are indistinguishable without resolving the pointer. The SDK's converter emits neither construct.

**Latent, not live — verified, not assumed.** A real stdio `tools/list` handshake against the rebuilt bundle returns a payload whose SHA-256 is **identical** before and after this commit (`fc73725965…e92fc4`), so no shipped schema was ever corrupted. `get-checklist-state`'s `outputSchema` has a `properties.items` that collides with the `items` keyword, but it lands in a safe branch. Fixed before release rather than after.

**Verification**
- Unit tests **557 → 565** (+8). **6 of the 8 fail against the pre-fix converter** — confirmed by restoring the old source and re-running. The other two are the coverage guard for the malformed-map branch and the `get-checklist-state` regression guard, which is *expected* to pass on both and is what establishes the defect was latent.
- Gate green: `lint`, `typecheck`, `test`, `format:check`, plus `test:coverage` (thresholds hold) and `test:integration` (the outputSchema contract suite, 5 tests, real server over stdio).
- Handshake re-confirmed on the committed bundle: **36 tools** (unchanged), all 36 `inputSchema` **and** all 36 `outputSchema` declaring `https://json-schema.org/draft/2020-12/schema`, **zero** `draft-07` substrings.
- `build/index.js` rebuilt, committed, and byte-identical to a fresh `pnpm run build`.

No version bump — this rides in the 2.7.2 release the PR already carries; the CHANGELOG entry was added under the existing `## [2.7.2]` heading.

PR #133, opened 2026-08-12
docs: remove the Testing Your Understanding self-quiz from CLAUDE.md

oliverames:
CLAUDE.md ended with a "Testing Your Understanding" section that reran three examples from the Critical: Backslash Escaping table as a quiz. The model reading the file already has those rules a few sections up, so the quiz spends context without adding information, and current guidance on agent context files flags exactly this pattern as removable.

This PR deletes the quiz section and nothing else. The escaping table, the "Why This Matters" explanation, and the worked examples all stay where they were, since those are the real gotchas.

The change only touches CLAUDE.md, but lint, the full test suite (565 tests), and the build all pass locally.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -211,14 +211,6 @@ This works in: `create-note` (folder param), `create-folder`, `search-notes`, `l
 | "iCloud sync in progress" | Wait and retry - results may be incomplete |
 | "No checklist items found" | Note has no checklists, or Full Disk Access not granted |
 
-## Testing Your Understanding
-
-Before creating notes with shell commands or paths containing backslashes, verify you're escaping correctly:
-
-- `~/path/to/file` - No escaping needed (no backslashes)
-- `Mobile\ Documents` - Needs escaping: `Mobile\\ Documents`
-- `C:\Users\` - Needs escaping: `C:\\Users\\`
-
 ## Recurring macOS permission prompts → offer the official-Node fix
 
 If a user reports being **repeatedly** prompted for Full Disk Access or


sweetrb:
Merged — thanks, @oliverames.

I checked the redundancy claim rather than taking it at face value, and it mostly holds: `Mobile\ Documents` was a verbatim repeat of the escaping table's third row, and `~/path/to/file` is derivable from the stated rule. A self-quiz that re-derives rules stated forty lines earlier is exactly the kind of thing that costs context without adding information, and you scoped it correctly — the table, the "Why This Matters" explanation and the worked examples are the parts that carry the actual gotchas, and you left all three alone.

One thing worth recording, since it came out of reviewing your diff and it's **not** a criticism of it: the quiz's third line was the one place in the file that got Windows paths *right*.

The quiz said `C:\Users\` → `C:\\Users\\`, which follows the table's rule (`\` → `\\`). But the Examples section says:

```
**Correct - Windows path:**
content: "Path: C:\\\\Users\\\\Documents"
```

Per that same table (`\\` → `\\\\`), the four-backslash form decodes to a literal `C:\\Users\\Documents` — doubled backslashes, not a Windows path. It also contradicts the shell example directly above it, which correctly uses `\\` for a single literal backslash.

So the file had a correct statement and an incorrect one, and removing the quiz happened to leave the incorrect one standing. That's a pre-existing bug in the Examples section, not something your PR introduced — your change is right either way. I'm fixing the example separately, along with a stale "Default account is iCloud" line that 2.7.1 (#129) invalidated when account resolution stopped hardcoding that name.

Follow-up: #134.

PR #144, opened 2026-08-29
fix: prevent unsafe Apple Notes mutations

artemgetmann:
## Summary

This closes three write-safety failures reproduced against real disposable Apple Notes:

- duplicate titles could make a title-based mutation target the wrong note
- a stale read could silently overwrite a newer edit
- full-body update or append could remove attachments

## Changes

- require canonical CoreData note IDs for update, append, delete, move, and batch move
- return a SHA-256 `contentHash` from exact note reads
- require `expectedContentHash` for update, append, delete, and batch delete
- compare the complete body and mutate in one no-retry AppleScript transaction
- reject update and append when attachments exist, including a second in-transaction check
- read the same note ID after body writes and report `verifiedVisibleText`
- verify the actual destination folder ID for single and batch moves
- fail closed when create-note does not return a canonical Apple Notes ID
- update schemas, README, skill copies, bundled runtime, plugin manifests, version, and changelog

## Contract changes

This intentionally removes title-only mutations. Titles remain available for search and reading.

`batch-delete-notes` now accepts reviewed `{ id, expectedContentHash }` snapshots instead of a bare ID array.

Apple Notes normalizes stored HTML. Post-save verification therefore proves the visible text and returns the exact post-save hash; it does not claim byte-identical rich formatting.

## Verification

- `pnpm test`: 601 passed
- `pnpm run lint`: passed
- `pnpm run format:check`: passed
- `pnpm run typecheck`: passed
- `pnpm run build`: passed
- `pnpm run test:integration`: 18 passed against Notes.app

Disposable live acceptance also verified:

- title-only update is rejected
- exact-ID update changes only the intended note when two notes share a title
- a stale save is rejected and the newer content survives
- attachment-bearing update and append are rejected while an attachment remains
- successful writes read back from the same ID
- single and batch moves verify and reach the requested folder

All disposable notes were moved to Recently Deleted after testing.


sweetrb:
Merged — thank you for this. The atomic-transaction design was exactly right: putting the body comparison, attachment check, and write inside one no-retry AppleScript closes the JS-side check-then-write race that a naive guard would leave open, and the disposable-note acceptance testing against real Notes.app (duplicate titles, stale writers, attachment rejection, destination-folder verification) is the kind of empirical validation this class of fix actually needs.

While reviewing I found one more gap in the same spirit as what you closed: AppleScript's `is`/`is not` string comparison is case-insensitive by default, so a case-only concurrent edit (e.g. autocorrect fixing capitalization) between a caller's read and the atomic write would read as "unchanged" and slip past the conflict guard — undermining the exact guarantee this PR advertises. I added `considering case ... end considering` around both comparisons (in `updateNoteByIdIfUnchanged` and `deleteNoteByIdIfUnchanged`) plus regression assertions pinning it, verified empirically that AppleScript's default comparison really is case-insensitive, and pushed it to your branch before merging (`maintainerCanModify` was on — thank you for leaving that open, it made landing this same-day possible). Full test suite, lint, typecheck, and format all green after the change.

Also filed #145 during review to track two narrow edge cases you and I both agree are worth documenting but not worth blocking this merge over (the linefeed-tolerant guard, and a possible false readback-mismatch from `comparableVisibleText`'s tag-boundary spacing) — not from your PR, from the design space this PR operates in.

This closes a real class of data-loss bugs (duplicate-title mistargeting, stale-write clobbering, attachment-dropping full-body rewrites) that had no guard before. Appreciated.

PR #155, opened 2026-09-10
fix(readback): don't invent a space at inline-tag boundaries (2.8.4)

### LINKED ISSUE #145: Two narrow false-signal edge cases in the #144 write-safety guards
Split out of review on #144 (\"fix: prevent unsafe Apple Notes mutations\") — not blocking that merge, since both are narrow edge cases with no test coverage either way and the wrong-direction fix risk (hand-modifying a safety-critical AppleScript string comparison without empirical Notes.app validation) outweighs leaving them documented for now.

## 1. The linefeed-tolerant conflict guard can mask a genuine one-linefeed edit

`updateNoteByIdIfUnchanged` / `deleteNoteByIdIfUnchanged` compare the current body against the expected body two ways:

```applescript
if currentBody is not "${safeExpectedBody}" and currentBody is not "${safeExpectedBody}" & linefeed then return "SAFETY_CONFLICT"
```

The second form exists because `executeAppleScript`'s `output.trim()` strips a real trailing linefeed Notes.app's `body` property can legitimately carry — without it, a same-content read would spuriously conflict with itself. But the same tolerance also matches a note that changed *concurrently* by exactly one trailing linefeed (added or removed, nothing else) between the caller's read and the write. In that narrow window, a genuinely-changed note is treated as unchanged and the write proceeds — the exact race this PR closes, reopened for one specific edit shape.

Likely low-frequency (requires a concurrent edit whose only difference is a single trailing linefeed), but worth a regression test once there's a reliable way to reproduce Notes.app's linefeed behavior deterministically, and worth reconsidering whether the trim() should move to a normalize-both-sides comparison instead of a two-literal OR.

## 2. `comparableVisibleText` can report a false readback-mismatch on a successful HTML write

The readback-verification normalizer (`src/utils/noteRevision.ts`) inserts a space at every tag boundary before collapsing whitespace, so immediately-adjacent inline tags produce a forced space that literal adjacent text wouldn't have (`<b>foo</b><i>bar</i>` → \"foo bar\", not \"foobar\"). If Apple Notes' save-time HTML normalization ever merges two adjacent inline runs without inserting a character between them, the readback's `comparableVisibleText` would drop that forced space while the just-written `writtenBody`'s transform keeps it — producing a spurious \"readback visible text did not match\" error (`update-note`/`append-to-note`) for a write that actually succeeded. Annoying (false alarm on a real success), not data-loss (wrong direction from #144's real concern), but worth a note or a test once someone can confirm whether Notes.app's save path actually merges adjacent formatting runs like that in practice.

Neither of these reproduced in #144's disposable-note or integration testing — recording them here so they're not lost, per this repo's usual practice of tracking known-unexplained-or-unconfirmed edge cases (see #155, #203) rather than shipping them silently undocumented.

sweetrb:
Both edge cases from #145, investigated against **real Notes.app** rather than reasoned about. One is a genuine bug and is fixed; the other turns out to be impossible.

## Fixed — the false readback mismatch

**Notes.app merges adjacent inline runs of the same style on save.** Verified 2026-09-10:

| written | stored by Notes |
|---|---|
| `<b>foo</b><b>bar</b>` | **`<b>foobar</b>`** |
| `<b>same</b><b>style</b><b>runs</b>` | **`<b>samestyleruns</b>`** |
| `<b>alpha</b><i>beta</i>` | unchanged (different styles) |

`comparableVisibleText` replaced *every* tag with a space, so the written side normalised to `merge me` and the readback to `mergeme`. `update-note`/`append-to-note` then failed a **successful** write with *"readback visible text did not match. Do not retry automatically; inspect note ID … in Notes.app"* — reproduced on 2.8.3, with the note confirmed correctly written.

Inline tags now collapse to nothing. That also matches rendering — `<b>foo</b><b>bar</b>` displays as `foobar` — so the old transform mis-described the **written** side too, not just the readback.

Scoped as an allow-list of tags non-separating by definition. `div`, `p`, `li`, headings, table cells, `<br>` and any unrecognised tag still separate words exactly as before, and `\b` stops the list eating `<bdo>`/`<summary>`. Both pre-existing equality tests still pass unchanged.

**Verified end to end**: the same input that errors on 2.8.3 returns `visible text verified` on this build, with Notes still storing the merged `<b>mergeme</b>`.

## Refuted — the linefeed guard cannot mask an edit

The concern was that tolerating `expected & linefeed` could hide a concurrent edit differing by exactly one trailing linefeed. It can't: **Notes normalises trailing linefeeds.**

| body set to | reads back as |
|---|---|
| `X` | `X\n` |
| `X\n` | `X\n` |
| `X\n\n` | `X\n` |

So no distinct "differs by one trailing linefeed" state exists for the tolerance to mask. The tolerance is also genuinely *required* — a real read always carries the linefeed while `expectedBody` is trimmed, so without it every unchanged read would self-conflict. The guard is sound; left unchanged. (The bare `expected` alternative is simply unreachable, which is harmless.)

## Gates
lint · typecheck · **598 tests** (5 new, verified to fail against the old transform) · prettier · bundle rebuilt

Fixes #145.


sweetrb:
### CodeQL `js/incomplete-multi-character-sanitization` — dismissed as a false positive

Recording the reasoning here, since the dismissal comment is capped at 280 characters.

**`comparableVisibleText` is a comparison normaliser, not a sanitiser.** It has exactly two call sites — `src/index.ts:973` and `src/index.ts:1133` — and both are `!==` equality checks whose only effect is to select an error branch:

```ts
if (!readback || comparableVisibleText(readback) !== comparableVisibleText(result.writtenBody)) {
  return errorResponse("… readback visible text did not match …");
}
```

The return value is never rendered, never written back to a note, never emitted as HTML, and never returned to a caller. There is no sink for the `<script` an incomplete strip could leave behind, which is the premise the rule reasons from.

**On the trigger itself.** The `<[^>]*>` pass predates this PR and was never flagged; what newly trips the rule is that a matched tag may now be replaced with an *empty string* rather than a space, which is the shape of the classic `<scr<script>ipt>` bypass. That bypass does not actually apply here: the global pass matches `<scr<script>` as a single tag, which is not on the inline allow-list, so it is replaced with a **space**, leaving `ipt>`. Only tags whose name is on the allow-list can collapse to nothing.

I restructured first rather than dismissing first — the two chained strip-regexes of the initial commit became a single `<[^>]*>` pass with a replacer (`def19b0`), restoring the exact regex shape the file already had. The alert persisted, which confirmed it is the empty-string replacement that trips the rule, not the chaining. Keeping the single-pass form regardless: it is simpler and puts the inline/block decision in one place.

PR #156, opened 2026-09-13
fix: preserve embedded note links and rich revisions [1/6]

NikoVonLas:
## Problem

Apple Notes can omit link destinations from AppleScript HTML. A client that reads and writes that body can therefore lose links or other native rich content without noticing.

## Changes in this PR

- Restore actual link destinations from the requested note's read-only rich-text metadata.
- Return `links`, `nativeTags`, `richContentComplete`, and `writable` from `get-note-content`.
- Include rich metadata in revision hashes so link-only and native-object changes invalidate stale writes.
- Refuse lossy full-body writes and require `allowLinkChanges` for intentional link changes.
- Verify retained links after updates and appends.

## PR series

This is part **1 of 6** in a review-ordered series:

1. [#156 — Preserve note links and rich revisions](https://github.com/sweetrb/apple-notes-mcp/pull/156) **(this PR)** — restore real URLs and prevent lossy rich-content writes. **Status:** merged.
2. [#157 — Inspect native note objects and tags](https://github.com/sweetrb/apple-notes-mcp/pull/157) — read native objects, checklists, tags, and table structure. **Status:** ready for review.
3. [#158 — Rename folders and add verified attachments](https://github.com/sweetrb/apple-notes-mcp/pull/158) — perform exact-ID folder and attachment mutations. **Status:** draft.
4. [#159 — Add a verified native tags bridge](https://github.com/sweetrb/apple-notes-mcp/pull/159) — create real Notes tags through a guarded Shortcut bridge. **Status:** draft.
5. [#160 — Add verified native background editing](https://github.com/sweetrb/apple-notes-mcp/pull/160) — append and edit supported native content without body replacement. **Status:** draft.
6. [#161 — Add an explicit Shortcut setup flow](https://github.com/sweetrb/apple-notes-mcp/pull/161) — install and diagnose the optional native-write bridges. **Status:** draft.

Review and merge the series in the listed order.

## Validation

- Lint, formatting, typecheck, skill-copy and plugin-manifest checks pass.
- `pnpm run test:coverage`: 617 tests across 26 files; configured thresholds pass.
- The committed bundle matches the source.


NikoVonLas:

sweetrb:
Merged, thanks @NikoVonLas — this is a genuinely careful piece of work.

A few things worth calling out specifically:

- `styleValue`'s field-9 paragraph-UUID stripping is exactly right: it walks the raw protobuf bytes to drop only the 16-byte paragraph identity field Notes regenerates on first native edit, while conservatively keeping every other field (including unknown ones) in the style signature. The test that proves this — same styleRuns, different revision hash across two synthetic paragraph UUIDs — is a great way to pin that distinction down.
- `restoreNoteLinks`'s token-based character alignment (matching visible characters between the AppleScript HTML and the DB's rich-text run positions, then reinserting `<a>` spans back-to-front by offset) handles adjacent/merged inline runs correctly without assuming any particular HTML shape from Notes.app's export.
- The real-note validation call-out in the PR body matches what I found independently — I ran `readRichNote` against a sample of my own live NoteStore.sqlite notes (read-only, no AppleScript) and it recovered `http://`, `https://`, and `mailto:` link destinations correctly, including a note where the note title itself carried a link and a checklist note, with zero exceptions.

One thing I did as part of merging: the committed `build/index.js` was stale relative to `src/index.ts` (332 lines out of date), which was silently failing CI's "Verify committed build/ matches source" step on both Node versions — that's also why the PR was showing zero passing checks going in, on top of the fork-approval gate. I rebuilt and pushed that fix directly to your branch since `maintainerCanModify` was enabled; everything's green now.

Looking forward to review-ordered PRs 2–6 once they're out of draft.

NikoVonLas:
Thanks for the detailed review and for rebuilding the bundle before merge. I traced the stale bundle to building from a worktree whose `node_modules` was symlinked from a neighboring checkout; esbuild then produced a different path-sensitive bundle. The remaining PR branches now use dependencies installed directly in their own worktree. #157 has been rebased onto v2.8.5, rebuilt with no diff, and revalidated on Node 22 and 24; I am moving it out of draft next.

PR #157, opened 2026-09-13
feat: inspect native note objects and tags [2/6]

NikoVonLas:
## Problem

Text hashtags and AppleScript HTML do not expose enough information to distinguish native Notes tags, checklists, embedded objects, and table structure safely.

## Changes in this PR

- Add read-only `get-native-objects` for exact object IDs, checklist state, style metadata, and decoded table cells.
- Add scoped `list-native-tags` with partial-read reporting.
- Decode native Notes tables conservatively, preserving row and column identities and reporting incomplete data instead of guessing.
- Expose the inspection tools through the server and document their Full Disk Access requirement.

## PR series and dependencies

2. [#157 — Inspect native note objects and tags](https://github.com/sweetrb/apple-notes-mcp/pull/157) **(this PR)** — ready for review.
3. [#158 — Rename folders and add verified attachments](https://github.com/sweetrb/apple-notes-mcp/pull/158) — ready for review; depends on #157.
4. [#159 — Add a verified native tags bridge](https://github.com/sweetrb/apple-notes-mcp/pull/159) — ready for review; depends on #157.
5. [#160 — Add verified native background editing](https://github.com/sweetrb/apple-notes-mcp/pull/160) — draft; depends on #159.
6. [#161 — Add an explicit Shortcut setup flow](https://github.com/sweetrb/apple-notes-mcp/pull/161) — draft; depends on #160.

After #157, #158 and #159 are independent review tracks and can be reviewed in parallel. The current package versions still impose merge order #158 → #159 → #160 → #161; if the tags track is merged first, it will be rebased and renumbered before merge.

This branch is rebased directly onto `main` after #156, so the PR diff contains only this layer.

## Validation

- Coverage passed: 620 tests across 27 files on Node 22 and Node 24.
- Lint, formatting, typecheck, integration, version guard, bundle boot, and CodeQL pass.


NikoVonLas:

NikoVonLas:
#156 has merged. This branch is now rebased directly onto v2.8.5 and contains only the native inspection layer. Its committed bundle matches a clean in-worktree build; lint, formatting, typecheck, integration tests, and 620 coverage tests pass on Node 22 and Node 24. Marking this PR ready for review.

NikoVonLas:
Series dependency update after splitting the branches more cleanly:

- #156 is merged; #157 is the current ready layer.
- After #157, #158 (folder rename / attachments) and #159 (native tags bridge) are independent and can be reviewed in parallel.
- #160 depends on #159, and #161 depends on #160.
- The current package versions still impose merge order #158 → #159 → #160 → #161. If the tags track is merged first, I will rebase and renumber it before merge.

Each PR description now contains the dependency graph, links, layer-only fork comparison, and current validation results. #158 and #159 currently have green CI and CodeQL.

sweetrb:
Reviewed — **merging**, with two notes that don't block.

Thank you for the dependency graph in each PR body and for versioning the stack correctly off `main` (2.8.6 → 2.8.7 → 2.8.8). That made this reviewable as a set instead of three unrelated diffs, and it caught a sequencing issue I'd otherwise have found the hard way (see #159).

**What I checked in the decoder.** `parseNoteTable` fails closed in the places that matter: `gunzipSync` capped at 16 MiB, entry and row×column guards at 100k, CRDT-deleted rows/columns skipped rather than misindexed, an ambiguous root rejected outright, and — the part I liked most — a cell containing `￼` throws instead of emitting text that would silently be wrong. "Report incomplete rather than guess" is the right call for a format you're reading by observation, and the code actually does it.

The fixture is a real artifact too: 978 bytes decoding to `Имя / Статус / Проба 🧭 / Готово`, so the happy-path test is a genuine round-trip over non-ASCII rather than a shape assertion.

**Note 1 — the headline safety claim is the untested one.** Three tests cover happy path, truncated input and garbage bytes. The branches that aren't covered are the ones the PR description sells: the `￼` embedded-cell rejection, the deleted-CRDT `continue`, the RTL reversal, and `roots.length !== 1`. For 125 lines of binary parsing over user data, that's thin — and the embedded-cell path in particular is the one whose regression would be silent (wrong cell text, no error). Worth a follow-up with a couple of hand-built fixtures; not a merge blocker given the code is correct as written.

**Note 2 — two O(n²) shapes, bounded but worth knowing.** `ids.forEach(... uuids.indexOf(id) ...)` in `ordered()` is quadratic in table dimensions, and the alias fixpoint loop is `O(aliases²)` worst case. Both are capped by the 100k guards so there's no practical exposure, but if table sizes ever grow those are the two spots.

Tool descriptions carry the full `Use when: / Returns: / Do not use when: / Safety:` structure and correctly call out the Full Disk Access requirement, and `parseNoteTable` is wrapped at the call site so a throw degrades the tool rather than failing the server.

PR #158, opened 2026-09-13
feat: rename folders and add verified attachments [3/6]

NikoVonLas:
## Problem

Folder renames and attachment insertion need exact object identity and conflict checks; name-based copy/delete flows can target the wrong folder or lose state.

## Changes in this PR

- Add `get-folder-by-id` and guarded `rename-folder` using exact folder ID, expected name, and expected parent ID.
- Preserve folder identity, notes, and descendants while rejecting stale metadata and sibling collisions.
- Add exact-ID `add-attachment` with note revision checks, bounded local-file validation, no automatic retry, and byte-for-byte readback verification.
- Open and validate attachment input through one descriptor, rejecting symlinks and avoiding a check/read race.

## PR series and dependencies

2. [#157 — Inspect native note objects and tags](https://github.com/sweetrb/apple-notes-mcp/pull/157) — ready for review.
3. [#158 — Rename folders and add verified attachments](https://github.com/sweetrb/apple-notes-mcp/pull/158) **(this PR)** — ready for review; depends on #157.
4. [#159 — Add a verified native tags bridge](https://github.com/sweetrb/apple-notes-mcp/pull/159) — ready for review; also depends on #157, independently of this PR.
5. [#160 — Add verified native background editing](https://github.com/sweetrb/apple-notes-mcp/pull/160) — draft; depends on #159.
6. [#161 — Add an explicit Shortcut setup flow](https://github.com/sweetrb/apple-notes-mcp/pull/161) — draft; depends on #160.

After #157, this PR and #159 can be reviewed in parallel. The current package versions still impose merge order #158 → #159 → #160 → #161.

[Review only this PR layer in the fork comparison](https://github.com/NikoVonLas/apple-notes-mcp/compare/upstream/pr-02-native-inspection...upstream/pr-03-direct-operations). It will be rebased onto `main` and marked ready after #157 merges.

## Validation

- Coverage passed: 632 tests across 29 files.
- Lint, formatting, typecheck, integration, version guard, bundle boot, and CodeQL pass.


NikoVonLas:

sweetrb:
Reviewed — **merging, no changes requested.** This is the strongest of the three.

**The write semantics are what sold it.** `rename-folder` takes the expected current name *and* parent and refuses on stale metadata or a sibling conflict — compare-and-swap rather than last-write-wins, which is the right shape for an API an assistant drives without seeing the folder tree. `add-attachment` is equally careful: a fresh rich revision required, 64 MiB ceiling through a private temp file, **never retries insertion**, and verifies both the pre-existing content and the fetched bytes afterward. Not retrying is the detail I'd have asked for if it weren't already there — a retried attachment insert is how you get silent duplicates, and it's exactly the failure that's invisible until a user notices their note has two copies of the same file.

**Test ratio is right too** — 194 lines of test against 490 of source, covering the rename guards and the direct-operation paths rather than just the happy case.

Tool descriptions carry the full structured format, and the `Safety:` lines are specific enough to actually constrain a caller ("requires the expected current name and parent", "never retries insertion") rather than the usual "use with care".

Merging after #157 so the stack lands in order.

PR #159, opened 2026-09-13
feat: add verified native tags bridge [4/6]

NikoVonLas:
## Problem

Inline `#hashtags` are searchable text, but they do not prove that Apple Notes registered native tag objects. AppleScript cannot create those objects directly.

## Changes in this PR

- Add `native-tags-status` and `add-native-tags` backed by a packaged signed Shortcut.
- Require an exact note ID, fresh revision, title, and unique existing scope text before mutation.
- Verify native tags, text, links, and existing native objects by exact-ID readback after the Shortcut runs.
- Treat timeouts as uncertain transport outcomes and never repeat the write automatically.
- Include build tooling and setup documentation for the Native Tags Shortcut.

## PR series and dependencies

1. [#156 — Preserve note links and rich revisions](https://github.com/sweetrb/apple-notes-mcp/pull/156) — merged.
2. [#157 — Inspect native note objects and tags](https://github.com/sweetrb/apple-notes-mcp/pull/157) — ready for review.
3. [#158 — Rename folders and add verified attachments](https://github.com/sweetrb/apple-notes-mcp/pull/158) — ready for review; also depends on #157, independently of this PR.
4. [#159 — Add a verified native tags bridge](https://github.com/sweetrb/apple-notes-mcp/pull/159) **(this PR)** — ready for review; depends on #157.
5. [#160 — Add verified native background editing](https://github.com/sweetrb/apple-notes-mcp/pull/160) — draft; depends on this PR.
6. [#161 — Add an explicit Shortcut setup flow](https://github.com/sweetrb/apple-notes-mcp/pull/161) — draft; depends on #160.

After #157, this PR and #158 can be reviewed in parallel. The current package versions still impose merge order #158 → #159 → #160 → #161; if this PR is merged first, it will be rebased and renumbered before merge.

[Review only this PR layer in the fork comparison](https://github.com/NikoVonLas/apple-notes-mcp/compare/upstream/pr-02-native-inspection...upstream/pr-04-native-tags). It will be rebased onto `main` and marked ready after #157 merges.

## Validation

- Coverage passed: 634 tests across 29 files.
- Lint, formatting, typecheck, integration, version guard, bundle boot, and CodeQL pass.

PR #160, opened 2026-09-13
feat: add verified native background editing [5/6]

NikoVonLas:
## Problem

Full-body AppleScript replacement cannot safely edit notes containing native tags, checklists, tables, links, or other rich objects.

## Changes in this PR

- Add a packaged Background Operations Shortcut and verified native append for plaintext, bounded Markdown, and semantic HTML.
- Add native checklist creation, table creation, pin changes, static note-link insertion, native tag removal, and explicit tag replacement across selected notes.
- Add `get-capabilities` so clients can distinguish implemented, installed, live-verified, and unavailable operations.
- Guard every mutation with an exact note ID, fresh revision, unique scope text, and post-write preservation checks; document that Shortcuts lookup should use plain words without punctuation, hashtags, or paths.
- Route supported appends for protected notes through the native bridge instead of replacing the body.

## PR series and dependencies

2. [#157 — Inspect native note objects and tags](https://github.com/sweetrb/apple-notes-mcp/pull/157) — merged.
3. [#158 — Rename folders and add verified attachments](https://github.com/sweetrb/apple-notes-mcp/pull/158) — merged.
4. [#159 — Add a verified native tags bridge](https://github.com/sweetrb/apple-notes-mcp/pull/159) — merged.
5. [#160 — Add verified native background editing](https://github.com/sweetrb/apple-notes-mcp/pull/160) **(this PR)** — ready; rebased directly onto current `main` after #159.
6. [#161 — Add an explicit Shortcut setup flow](https://github.com/sweetrb/apple-notes-mcp/pull/161) — draft; depends on this PR.

The PR diff now contains only this layer. Its committed bundle was rebuilt after the rebase and includes the tools merged through #159 as well as this PR's native editing surface.

## Validation

- Coverage passed: 702 tests across 36 files.
- Integration passed: 13 tests, with 5 environment-dependent tests skipped.
- Lint, formatting, typecheck, four Shortcut fixture tests, version guard, and bundle boot pass.
- The rebuilt bundle contains `rename-folder`, `native-tags-status`, and `append-native`.
- CodeQL's two filesystem-race findings are resolved and outdated; both file paths now use one descriptor from inspection through read.


NikoVonLas:

NikoVonLas:
Rebased directly onto current main after #159 and rebuilt the committed bundle from the combined source. The bundle now includes the merged rename/attachment and native-tags tools plus this native-editing layer. Local validation: 702 coverage tests, 13 integration tests, lint, format, typecheck, Shortcut fixture tests, and explicit bundle symbol checks all pass.

sweetrb:
Reviewed — **merging, no changes requested.**

Thank you for rebasing this onto the merged `main` before marking it ready. I checked the thing that bit #159: this branch's `build/index.js` carries **both** `rename-folder` (#158) and the native-tags surface (#159), so the stale-bundle problem doesn't recur here.

**I verified the Shortcut the same way as #159, and it holds.** The committed unsigned artifact reproduces **byte-for-byte** from `scripts/build-native-operations-shortcut.py` (it composes the v1 tags builder already on `main`, which is why the reproduction needs both scripts present). Shipping a `test-native-operations-shortcut.py` alongside the builder is a genuine improvement over the previous round — the artifact is now not just auditable but exercisable.

**Test ratio is the strongest in the series:** ~890 lines of new source against ~661 of tests, and they're split across the API surface, the runner, and the background-notes service rather than concentrated on the happy path.

**The `Safety:` clauses are doing real work**, which matters more here than in #157/#158 because these are mutations against live notes:

- *"never substitutes text; preserves existing rich content and reports failure unless the native table is verified"*
- *"preserves existing text, links, native objects, checklist identities, and formatting"*
- *"requires a fresh revision and unique existing scope phrase of plain words; avoid punctuation, hashtags, and paths because Notes search may not resolve them literally"*

That last one is the kind of constraint an assistant can actually act on — it names the failure mode (Notes search not resolving a literal) rather than just warning to be careful. Readback-verified writes with an explicit fresh-revision requirement is the right shape for this surface.

Merging at 2.8.9. #161 is stacked on this, so it'll need the same rebase-and-rebuild once this lands — same as before, and you've clearly got the pattern.

PR #178, opened 2026-09-18
test: repair live integration tests that could never run

AJRC-Star:
## Problem

The `live Notes.app operations` group in `test/integration.test.ts` calls `mgr.deleteNoteById()`, which does not exist on `AppleNotesManager`. The only by-id delete is `deleteNoteByIdIfUnchanged(id, expectedBody)` — a guarded delete that takes a second argument and returns `{ status }` rather than a boolean.

The `beforeAll` probe wraps its work in a `try` whose `catch` is commented *"account not writable — try the next one"*. Instead of catching an AppleScript failure, it swallows the `TypeError` from the missing method. `liveAccount` stays `null`, all five live tests skip themselves, and the suite reports green.

Two consequences:

1. **These tests cannot ever have passed.** On a macOS runner with a working Notes account, `pnpm run test:integration` reports `13 passed | 5 skipped` and exits 0 — the skip is invisible unless you read verbose output.
2. **Each run leaks a note.** `createNote` succeeds, the delete throws, and `__mcp_probe__` is left in the user's library. Three runs on my machine left three of them behind.

Reproduced directly against the manager:

```
accounts: [{"name":"iCloud",...}]
createNote -> {"id":"x-coredata://…/ICNote/p2124","title":"__mcp_probe__",...}
THREW for iCloud : mgr.deleteNoteById is not a function
```

## Fix

- Add a `deleteNote()` helper that reads the current body back via `getNoteContentById()` and hands it to the guarded delete, returning its status.
- Narrow the swallowing `catch` to `createNote()` alone, and rethrow `TypeError` / `ReferenceError` so a programming error can no longer masquerade as an unwritable account.
- Scope the outer `try` to `listAccounts()`, so a headless runner with no Notes account still self-skips while nothing downstream is silently eaten.
- Treat a failed probe cleanup as a hard error naming the note and id, rather than leaving it in a real library and skipping on.

## Verification

On macOS 27, Node 26, iCloud account:

| | Before | After |
|---|---|---|
| `pnpm run test:integration` | 13 passed, 5 skipped | **18 passed** |

The five live tests now genuinely exercise Notes.app (479ms–1411ms each, previously 0ms). No notes are left behind — the probe and both per-test notes land in Recently Deleted as intended.

Also green: `pnpm run lint`, `pnpm run typecheck`, `pnpm run format:check`, `pnpm test` (795 passed), `pnpm run build`.

Test-only change, so per CONTRIBUTING it is exempt from the version bump and the committed bundle is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sweetrb:
Thanks for the sharp diagnosis, @AJRC-Star — catching that `deleteNoteById` doesn't exist anywhere in `AppleNotesManager` (only the guarded `deleteNoteByIdIfUnchanged`), and that the `try/catch` around the probe was silently swallowing that `TypeError` instead of the AppleScript failure it claimed to handle. That's a real bug: the five live tests could never have passed, and every run leaked a `__mcp_probe__` note into the library. This PR closed itself within seconds of opening with no explanation, so I carried the fix forward as #179 (your commit, same authorship, now with Co-authored-by credit) and merged it — CI green across the board. Appreciated.

